PR #6, opened 2018-07-06
fixed TaoType based on petsc master

haplav:
still support current petsc maint


jkruzik (changes requested):
Works OK.

I just published the next branch; please rebase this on top of next. Then you can do no-ff merge to next and no-ff merge next to master.

We'll need to discuss how (if) to support PETSc master. I'll raise this on the mailing list.

haplav:
Merging after email discussion

PR #7, opened 2018-07-06
Build: update gmakefile & gmakegen.py based on current PETSc versions.

haplav:
- fix macOS clang linker error `ld: malformed 32-bit x.y.z version number: -current_version`
- infer library version from the last git tag
- include gmakefile.test
  - reuse some definitions like PETSc's gmakefile does
  - step towards use of PETSc test harness


jkruzik (commented):
looks fine to me, but again please rebase on top of next before merging

PR #8, opened 2018-08-31
Feature qpc

jkruzik:
Introduces QPC (Quadratic Programming Constraints) interface and box constraint implementation.

Main features:
- QPC interface - interface for getting feasible step lengths, gradient splitting, etc.
- QPCBox - implementation of box and bound constraints
- Constraints on a subset of the solution components can be specified by IS

Joint work with @haplav. Developed from code initially contributed by @lukaspospisil.


jkruzik:
needs to be checked with Valgrind (after I fix my installation - https://sourceware.org/git/gitweb.cgi?p=valgrind.git;a=commitdiff;h=21a01b13e259b9a43f10f0046b2b3f409c11ea75)

Issue #16, opened 2020-04-28
Use PETSc's MatNullSpace to represent null space.


haplav:
Related to #13 .

jkruzik:
I do not like the idea. MatNullSpace is just an array of vectors which does not have any basic Mat operations (e.g., MatMult). We are mostly using null space matrices to multiply vecs/mats.

haplav:
OK, I was thinking about converting to a matrix if needed. But for the time being it's probably not worth the effort. There are more important things. Closing.

haplav:
After all, I think it's still worth considering.
* We could have a routine which converts it to MATDENSE cheaply. And MatMat operations internally work with columns of R anyway.
* NullSpace would be naturally attached to the operator.
* The nullspace computation using MUMPS could be contributed to PETSc directly. The code would probably even become simpler. But then use of MatNullSpace would be required to fit into PETSc.

haplav:
At the very least, we could start by attaching the nullspace to the operator directly using PetscObjectCompose. It shouldn't be a property of a QP nor MATINV.

haplav:
> It shouldn't be a property of a QP nor MATINV.

See also #30.

jkruzik:
> We could have a routine which converts it to MATDENSE cheaply.

When would you convert it? Would you store it? If so where? May I, suggest storing it in qp->R? ;)
You can have a sparse null space.
> And MatMat operations internally work with columns of R anyway.

Not when both matrices are of the same type e.g., AIJ or dense. Plain matvecs do not work with columns directly.
> The nullspace computation using MUMPS could be contributed to PETSc directly.

Sure, there is MatNullSpaceCreateRigidBody() so there could be something like MatNullSpaceCreateWithMumps().
> The code would probably even become simpler.

The code in PETSc yes but we would need a wrapper over such routine to handle our MatTypes (block diag).
> But then use of MatNullSpace would be required to fit into PETSc.

For the PETSc routine yes, but that does not mean we would need to use MatNullSpace in PERMON.

> At the very least, we could start by attaching the nullspace to the operator directly using PetscObjectCompose. It shouldn't be a property of a QP nor MATINV.

I do not see why it is better to have R attached to Mat operator instead of it being a pointer in the QP struct. However, I do see quite a few disadvantages:
- When and how would you destroy R if it is computed by PERMON?
- You would need PetscObjectQuery() in each function you wanted to use R which means a for loop with Strcmp().
[...]

PR #12, opened 2020-04-24
Allow QPTDualize with only lin. inequality

jkruzik:
\+ fix unused variable warnings in nest

PR #20, opened 2020-04-29
WIP: QPTDualize: improve support for non-singular Hessian

### LINKED ISSUE #13: QPTDualize with non-singular Hessian
1. if QP does not have a nullspace matrix QPTDualize will try to compute one with MUMPS
2. if empty nullspace matrix is attached/computed by QPTDualize the dual problem has an empty (unnecessary) linear  equality constraint

cc: @haplav @lukaspospisil

haplav:
Fix #13.

### DIFF
--- a/src/qp/interface/qptransform.c
+++ b/src/qp/interface/qptransform.c
@@ -788,11 +788,13 @@ static PetscErrorCode QPTDualizePostSolve_Private(QP child,QP parent)
     Vec f          = parent->b;
     Vec u          = parent->x;
     Vec tprim      = parent->xwork;
-    PetscBool flg;
+    PetscBool flg, spd;
 
     PetscFunctionBegin;
     TRY( PetscObjectQuery((PetscObject)F,"Kplus",(PetscObject*)&Kplus) );
     FLLOP_ASSERT(Kplus,"Kplus != NULL");
+    TRY( MatGetOption(parent->A,MAT_SPD,&spd) );
+    if (!spd && !alpha) SETERRQ(PetscObjectComm((PetscObject)parent), PETSC_ERR_PLIB, "parent->A not marked as SPD but no alpha present");
 
     /* copy lambda back to lambda_E and lambda_I */
     if (parent->BE && parent->BI) {
@@ -816,19 +818,21 @@ static PetscErrorCode QPTDualizePostSolve_Private(QP child,QP parent)
     TRY( VecAYPX(tprim, -1.0, f) );
     TRY( MatMult(Kplus, tprim, u) );
 
-    TRY( VecIsInvalidated(alpha,&flg) );
-    if (flg) {
-      /* compute alpha = (G*G')\G(G'*alpha) where G'*alpha = child->Bt_lambda */
-      TRY( VecIsInvalidated(child->Bt_lambda,&flg) );
+    if (alpha) {
+      TRY( VecIsInvalidated(alpha,&flg) );
       if (flg) {
-        TRY( MatMultTranspose(child->B,child->lambda,child->Bt_lambda) );
+        /* compute alpha = (G*G')\G(G'*alpha) where G'*alpha = child->Bt_lambda */
+        TRY( VecIsInvalidated(child->Bt_lambda,&flg) );
+        if (flg) {
+          TRY( MatMultTranspose(child->B,child->lambda,child->Bt_lambda) );
+        }
+        TRY( QPPFApplyHalfQ(child->pf,child->Bt_lambda,alpha) );
       }
-      TRY( QPPFApplyHalfQ(child->pf,child->Bt_lambda,alpha) );
-    }
 
-    /* u = u - R*alpha */
-    TRY( MatMult(parent->R, alpha, tprim) );
-    TRY( VecAXPY(u, -1.0, tprim) );
+      /* u = u - R*alpha */
+      TRY( MatMult(parent->R, alpha, tprim) );
+      TRY( VecAXPY(u, -1.0, tprim) );
+    }
     PetscFunctionReturn(0);
 }
 
@@ -912,6 +916,7 @@ PetscErrorCode QPTDualize(QP qp,MatInvType invType,MatRegularizationType regType
   PetscBool        B_nest_extension = PETSC_FALSE;
   PetscBool        mp = PETSC_FALSE;
   PetscBool        true_mp = PETSC_FALSE;
+  PetscBool        spd = PETSC_FALSE;
 
   PetscFunctionBeginI;
   PetscValidHeaderSpecific(qp,QP_CLASSID,1);
@@ -919,6 +924,8 @@ PetscErrorCode QPTDualize(QP qp,MatInvType invType,MatRegularizationType regType
   TRY( QPTransformBegin(QPTDualize, QPTDualizePostSolve_Private,NULL, QP_DUPLICATE_DO_NOT_COPY,&qp,&child,&comm) );
   TRY( QPAppendOptionsPrefix(child,"dual_") );
 
+  G = NULL;
+  e = NULL;
   Kplus_orig = NULL;
   B = NULL;
   c = qp->c;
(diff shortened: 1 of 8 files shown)


haplav:
@jkruzik I think I have it, modulo `tutorials-ex4_1_nsize-3_dual-1+assume_spd-0` failing due to `Nonconforming object sizes` in `QPCheckNullSpace()`. I will try to fix it today.

haplav:
And for dual
```
r = ||A*x - b + B'*lambda|| = 1.00e+00    rO/||b|| = 3.30e+01
```
so similar problem as in https://github.com/permon/permon/pull/12#issuecomment-620627680.

However, comparing `-dual 0` and `-dual 1`, it looks like the resulting x is the same
```
# -dual 1
QP Object: QP_0x84000000_6 1 MPI processes
  type: QP
  #0 in chain, derived by 
|| x|| = 9.89949494e+00    max( x) = 1.00e+00 =  x(5)    min( x) = -8.02e-12 =  x(99)    21674b0
|| b|| = 3.03014843e-02    max( b) = 0.00e+00 =  b(0)    min( b) = -3.06e-03 =  b(1)    216b180
||cE|| = 9.89949494e+00    max(cE) = 1.00e+00 = cE(1)    min(cE) = 0.00e+00 = cE(0)    216e150
r = ||A*x - b + B'*lambda|| = 1.00e+00    rO/||b|| = 3.30e+01
r = ||BE*x-cE||          = 7.34e-11    r/||b|| = 2.42e-09
```
```
# -dual 0
QP Object: QP_0x84000000_4 1 MPI processes
  type: QP
  #0 in chain, derived by 
|| x|| = 9.89949494e+00    max( x) = 1.00e+00 =  x(2)    min( x) = 0.00e+00 =  x(0)    11104b0
|| b|| = 3.03014843e-02    max( b) = 0.00e+00 =  b(0)    min( b) = -3.06e-03 =  b(1)    1114180
||cE|| = 9.89949494e+00    max(cE) = 1.00e+00 = cE(1)    min(cE) = 0.00e+00 = cE(0)    1117150
r = ||A*x - b + (B'*lambda)|| = 2.57e-13    rO/||b|| = 8.47e-12
r = ||BE*x-cE||          = 1.35e-12    r/||b|| = 4.45e-11
```
So it seems the problem is in KKT evaluation itself or in multipliers `lambda`.

jkruzik:
You are too quick! ;)

I opened the issue #13 to discuss how to approach the fix. I am reluctant to make the equivalence of SPD = non-singular. Admittedly, most of the library probably won't work with indefinite Hessian.

Moreover, there are parameters MatInvType and MatRegularizationType. It is questionable whether to have these as parameters or options, but if we have regularization, we could regularize semi-definite Hessian before proceeding with the actual dualization.

I thought that the solution could be either an option (e.g. "-qpt_dualize_nonsingular 1") or better yet have QPSetOperatorNullSpace() allow setting NULL but setting a flag as well. We can then determine if user wants us to compute the null space (QPSetOperatorNullSpace was used to set NULL) or not. The second approach is somewhat similar to the solution of singular systems in PETSc [Manual - section 4.6](https://www.mcs.anl.gov/petsc/petsc-current/docs/manual.pdf#section.4.6).

haplav:
> You are too quick! ;)

I find discussion above an actual code change more efficient :-)

> I opened the issue #13 to discuss how to approach the fix. I am reluctant to make the equivalence of SPD = non-singular. Admittedly, most of the library probably won't work with indefinite Hessian.

It's not equivalence, but definitely SPD implies non-singular. If it ever will be important, we can add `MatOption=MAT_NONSINGULAR` to PETSc. But currently PERMON assumes SPS, I believe.

> Moreover, there are parameters MatInvType and MatRegularizationType. It is questionable whether to have these as parameters or options, but if we have regularization, we could regularize semi-definite Hessian before proceeding with the actual dualization.

The reason why there is no `QPTRegularize()` or so which would be called before `QPTDualize()` is that the regularization is an indivisible part of forming the pseudoinverse. It makes no sense on its own. Applying the regularization to the primal Hessian yields a non-equivalent QP problem.

> I thought that the solution could be either an option (e.g. "-qpt_dualize_nonsingular 1") or better yet have QPSetOperatorNullSpace() allow setting NULL but setting a flag as well. We can then determine if user wants us to compute the null space (QPSetOperatorNullSpace was used to set NULL) or not. The second approach is somewhat similar to the solution of singular systems in PETSc Manual - section 4.6.

* I still think using `MatOption` from PETSc is the most natural way. That said, `-qpt_dualize_nonsingular` can be an alternative way of setting it.
* `QPSetOperatorNullSpace()` should indeed be fixed to allow NULL. Easy fix, I just did it. But it's NULL by default anyway.
* Within this MR, if the user passes no nullspace and does not specify the flag, the null space is computed automatically and set to NULL (and flag set) if it has no columns. I think this is convenient behavior.
* I think the nullspace should be attached to the matrix, not the QP. It's awkward that it's a separate object in QP although it's basically a property of the operator. But that's a different story.

haplav:
`MatInvComputeNullSpace_Inv()` obeys the SPD flag already. So I will just change it not to generate a zero-sized matrix but NULL, instead of doing that in `QPTDualize()`.

jkruzik:
> > You are too quick! ;)
> 
> I find discussion above an actual code change more efficient :-)
> 
> > I opened the issue #13 to discuss how to approach the fix. I am reluctant to make the equivalence of SPD = non-singular. Admittedly, most of the library probably won't work with indefinite Hessian.
> 
> It's not equivalence, but definitely SPD implies non-singular. If it ever will be important, we can add `MatOption=MAT_NONSINGULAR` to PETSc. But currently PERMON assumes SPS, I believe.
> 
Yes, PERMON assumes SPS, but only implicitly - using SPD flag to indicate non-singularity makes the assumption explicit.
> > Moreover, there are parameters MatInvType and MatRegularizationType. It is questionable whether to have these as parameters or options, but if we have regularization, we could regularize semi-definite Hessian before proceeding with the actual dualization.
> 
> The reason why there is no `QPTRegularize()` or so which would be called before `QPTDualize()` is that the regularization is an indivisible part of forming the pseudoinverse. It makes no sense on its own. Applying the regularization to the primal Hessian yields a non-equivalent QP problem.
> 
It may or may not be equivalent (see e.g., Friedlander, Tseng - Exact Regularization of Convex Programs), but from what I gather solving a regularized problem is useful even when it is not exactly equivalent to solving the original problem.
> > I thought that the solution could be either an option (e.g. "-qpt_dualize_nonsingular 1") or better yet have QPSetOperatorNullSpace() allow setting NULL but setting a flag as well. We can then determine if user wants us to compute the null space (QPSetOperatorNullSpace was used to set NULL) or not. The second approach is somewhat similar to the solution of singular systems in PETSc Manual - section 4.6.
> 
>     * I still think using `MatOption` from PETSc is the most natural way. That said, `-qpt_dualize_nonsingular` can be an alternative way of setting it.
> 
>     * `QPSetOperatorNullSpace()` should indeed be fixed to allow NULL. Easy fix, I just did it. But it's NULL by default anyway.
> 
>     * Within this MR, if the user passes no nullspace and does not specify the flag, the null space is computed automatically and set to NULL (and flag set) if it has no columns. I think this is convenient behavior.

Should we, by default, assume that the Hessian is singular or non-singular?
> 
>     * I think the nullspace should be attached to the matrix, not the QP. It's awkward that it's a separate object in QP although it's basically a property of the operator. But that's a different story.
I actually prefer to keep it as it is :)

haplav:
> @jkruzik I think I have it, modulo `tutorials-ex4_1_nsize-3_dual-1+assume_spd-0` failing due to `Nonconforming object sizes` in `QPCheckNullSpace()`. I will try to fix it today.

The problem with `QPCheckNullSpace()` is a bit deeper than I thought. The problem is in `MatInvComputeNullSpace_Inv` which essentially supports only `MATBLOCKDIAG`. I'm working on generalization. I think it will be a separate pull request.

haplav:
TODO:

- [ ] in ex3.c, remove
  ```
    ierr = PetscOptionsInsertString(NULL,"-qpt_dualize_B_nest_extension 0 -qpt_dualize_G_explicit 0");CHKERRQ(ierr); /* workaround for empty nullspace */
  ```
  (https://github.com/permon/permon/pull/12#discussion_r417830837)

jkruzik:
Maybe we can also swap ex3 and ex4. I think it is more natural to start with linear equality constraint.

haplav:
> Maybe we can also swap ex3 and ex4. I think it is more natural to start with linear equality constraint.

It is, but let's keep it like this until they are both merged, and then swap them in a separate minor pull request. It would cause me headaches to swap them right away.

haplav:
TODOs:

- [ ] rebase after #23 gets merged
- [ ] in ex3.c, remove
  ```
    ierr = PetscOptionsInsertString(NULL,"-qpt_dualize_B_nest_extension 0 -qpt_dualize_G_explicit 0");CHKERRQ(ierr); /* workaround for empty nullspace */
  ```
  (https://github.com/permon/permon/pull/12#discussion_r417830837)
- [ ] update ex4 from ex3
- [ ] swap ex3/ex4

jkruzik:
What is the status on this PR?

haplav:
This will need rebasing and reworking. I defer it until prerequisites are resolved:
* [x] PR #23
* [ ] New PR which will move nullspace to Mat as discussed in Issue #16.

I hope for finishing in a week from now.

haplav:
TODO
- [x] New PR which will move nullspace to Mat as discussed in Issue #16.
- [x] rebase onto master once #23 is merged
- [x] change base back to `master`
- [x] in ex3.c, remove
  ```
    ierr = PetscOptionsInsertString(NULL,"-qpt_dualize_B_nest_extension 0 -qpt_dualize_G_explicit 0");CHKERRQ(ierr); /* workaround for empty nullspace */
  ```
  (https://github.com/permon/permon/pull/12#discussion_r417830837)
- [ ] finish #47 first
- [ ] update ex4 from ex3
- [ ] swap ex3/ex4
- [ ] if empty R is generated, maybe it could be just NULL rather than empty matrix, see https://github.com/permon/permon/pull/23#discussion_r431976425
- [ ] we need to make sure R is recomputed if K has changed - see https://github.com/permon/permon/pull/47#discussion_r442145954

haplav:
PR #47 will precede this one.

jkruzik:
stale and mostly reimplemented by #58

PR #28, opened 2020-05-05
gmakefile.test: Fix for multiple test directories.

haplav:
Update and simplify.
Add src/tests/ex1.c


gmakefile.test
```diff
@@ -263,7 +267,7 @@ else ifdef searchin
 else ifdef argsearch
   TESTTARGETS := $(foreach v,$(alltesttargets),$(if $(findstring $(argsearch),$($(v)_ARGS)),$(v)))
 else ifdef globsearch
-  TESTTARGETS := $(shell $(PYTHON) -c"import sys,fnmatch,itertools; m=[fnmatch.filter(sys.argv[2].split(),p) for p in sys.argv[1].split()]; print(' '.join(list(itertools.chain.from_iterable(m))))" '$(globsearch)' '$(alltesttargets)')
+  TESTTARGETS := $(shell $(PYTHON) $(PETSCCONFIGDIR)/query_tests.py 'name' '$(globsearch)')
```
jkruzik: This won't work with PETSc 3.13.0.
On the other hand, it is always possible to use search in our case.

jkruzik (approved):

gmakefile.test
haplav: @jkruzik I reverted this but added a comment. I also added a few more minor changes. Please look again. Please test it on your end.

Note it didn't work for me even with master. I don't wanna spend even more time by figuring out what's the problem. Let's look at it again at some point in the future.

In my ideal world, PETSc test harness should improve support for external packages like PERMON or SLEPc so that we don't have to maintain this whole file but include some PETSc one instead.

gmakefile.test
```diff
@@ -83,13 +88,8 @@ ifdef TIMEOUT
   TESTFLAGS += -t $(TIMEOUT)  # Override the default timeout
 endif
 
-PETSCCONFIGDIR ?= $(PETSC_DIR)/config
-ifeq ($(wildcard $(PETSCCONFIGDIR)/gmakegentest.py),)
-  PETSCCONFIGDIR := $(PETSC_DIR)/share/petsc/examples/config
-endif
```
jkruzik, line 89: Does not removing this break prefix installs?

gmakefile.test
```diff
@@ -101,8 +101,8 @@ else
   cc_name := CC
 endif
 
-PETSC_COMPILE.c = $(call quiet,$(cc_name)) -c $(PCC_FLAGS) $(CFLAGS) $(CCPPFLAGS) $(C_DEPFLAGS)
-PETSC_COMPILE.cxx = $(call quiet,CXX) -c $(CXX_FLAGS) $(CFLAGS) $(CCPPFLAGS) $(CXX_DEPFLAGS)
+PETSC_COMPILE.c = $(call quiet,$(cc_name)) -c $(PCC_FLAGS) $(PFLAGS) $(CCPPFLAGS) $(C_DEPFLAGS)
+PETSC_COMPILE.cxx = $(call quiet,CXX) -c $(CXX_FLAGS) $(CXXFLAGS) $(CXXCPPFLAGS) $(CXX_DEPFLAGS)
```
jkruzik, line 106: in lib/permon/conf/permon_variables, there shoulde be CXXCPPFLAGS = CCPPFLAGS

gmakefile.test
```diff
@@ -263,19 +263,23 @@ else ifdef searchin
...
 else ifdef globsearch
+  # PETSc uses
+  # TESTTARGETS := $(shell $(PYTHON) config/query_tests.py 'name' '$(globsearch)')
+  # since v3.13.1 but it is currently not working for us
   TESTTARGETS := $(shell $(PYTHON) -c"import sys,fnmatch,itertools; m=[fnmatch.filter(sys.argv[2].split(),p) for p in sys.argv[1].split()]; print(' '.join(list(itertools.chain.from_iterable(m))))" '$(globsearch)' '$(alltesttargets)')
 else ifdef test-fail
   TESTTARGETS := $(shell $(PETSC_ARCH)/tests/echofailures.sh)
 else ifdef query
-  TESTTARGETS := $(shell $(PYTHON) config/query_tests.py '$(query)' '$(queryval)')
+  TESTTARGETS := $(shell $(PYTHON) $(PETSCCONFIGDIR)/query_tests.py '$(query)' '$(queryval)')
```
jkruzik:
```suggestion
  TESTTARGETS := $(shell $(PYTHON) $(PETSCCONFIGDIR)/query_tests.py --testdir=$(TESTDIR) --srcdir=$(TESTSRCDIR) '$(query)' '$(queryval)')
```
otherwise it fails

gmakefile.test
```diff
@@ -263,19 +263,23 @@ else ifdef searchin
 else ifdef argsearch
   TESTTARGETS := $(foreach v,$(alltesttargets),$(if $(findstring $(argsearch),$($(v)_ARGS)),$(v)))
 else ifdef globsearch
+  # PETSc uses
+  # TESTTARGETS := $(shell $(PYTHON) config/query_tests.py 'name' '$(globsearch)')
+  # since v3.13.1 but it is currently not working for us
```
jkruzik:
```suggestion
  # TODO v3.14
  # TESTTARGETS := $(shell $(PYTHON) config/query_tests.py --testdir=$(TESTDIR) --srcdir=$(TESTSRCDIR) 'name' '$(globsearch)')
```
the original doesn't work only with a large number of tests (not our case) - we will switch to the update query_tests with 3.14

haplav: I meant it just didn't work but that was probably because of the missing `--testdir` and `--srcdir `- will try again.

haplav: If it is working like this with PETSc v3.13.1, we should already use this way. I don't think it's worth the TODO (and liability). I think the user who is interested in running tests will definitely be able to get the latest version.

gmakefile.test
haplav: ok, will do

gmakefile.test
haplav: aha ok, I though it was some old version legacy - will revert

haplav: But shouldn't we just check for empty PETSC_ARCH?

jkruzik: Empty PETSC_ARCH is a different use case from prefix install. With empty PETSC_ARCH gmakegentest is still in $(PETSC_DIR)/config.

gmakefile.test
haplav: OK, the new form works good with v3.13.1, so I pushed it.

gmakefile.test
haplav: done in 10c51ec

gmakefile.test
```diff
@@ -263,19 +263,20 @@ else ifdef searchin
 else ifdef argsearch
   TESTTARGETS := $(foreach v,$(alltesttargets),$(if $(findstring $(argsearch),$($(v)_ARGS)),$(v)))
 else ifdef globsearch
-  TESTTARGETS := $(shell $(PYTHON) -c"import sys,fnmatch,itertools; m=[fnmatch.filter(sys.argv[2].split(),p) for p in sys.argv[1].split()]; print(' '.join(list(itertools.chain.from_iterable(m))))" '$(globsearch)' '$(alltesttargets)')
+  TESTTARGETS := $(shell $(PYTHON) $(PETSCCONFIGDIR)/query_tests.py --testdir=$(TESTDIR) --srcdir=$(TESTSRCDIR) 'name' '$(globsearch)')
```
jkruzik, line 267: I am bit reluctant to unnecessary break the functionality with 3.13.0 when the original solution will work perfectly OK (for us) even with 3.13.1. On the other hand, I don't think the globsearch is actually really useful (for us).

gmakefile.test
haplav: I wonder what does empty PETSC_ARCH with no prefix install even mean?

jkruzik: There is no PETSC_ARCH with prefix installs. It is primarily used in packages and modules. See. https://www.mcs.anl.gov/petsc/documentation/installation.html#root

haplav: My point is what is the difference between "empty PETSC_ARCH" and "prefix install" case. As you just wrote, there is no PETSC_ARCH with prefix installs. So why not consider "empty PETSC_ARCH" implying the "prefix install" case.

Is it possible to have non-prefix PERMON on prefix PETSc and vice-versa?

jkruzik: Empty PETSC_ARCH does not imply prefix install. Prefix install implies empty PETSC_ARCH. If I remember correctly you can have installs with:
PETSC_DIR+PETSC_ARCH - lib is in $PETSC_DIR/$PETSC_ARCH/lib/
PETSC_DIR+ empty PETSC_ARCH - lib is in $PETSC_DIR/lib/?
prefix - lib is in $prefix/lib ? and lot of makefiles and scripts are somewhere under $prefix and not in the same location as in the standard installs

PERMON does not support prefix - there is no configure.

gmakefile.test
haplav: I would be reluctant if it was a change in master but it's a change in maint/patch release. This approach simpler and keeps pace with PETSc. Yes, globsearch is probably not that much used - the more so we can take the newer approach, because probability that somebody would need it with older PETSc version is ~0.

gmakefile.test
haplav: > PETSC_DIR+ empty PETSC_ARCH - lib is in $PETSC_DIR/lib/?
```
$ ./configure PETSC_ARCH=""
===============================================================================
             Configuring PETSc to compile on your system                       
===============================================================================
TESTING: configureArchitecture from PETSc.options.arch(config/PETSc/options/arch.py:43)                                                                                                                             *******************************************************************************
         UNABLE to CONFIGURE with GIVEN OPTIONS    (see configure.log for details):
-------------------------------------------------------------------------------
PETSC_ARCH cannot be empty string. Use a valid string or do not set one. Currently set with option PETSC_ARCH=
*******************************************************************************
```
So your option 2 does not exist, and I think we are safe to assume that  `PETSC_ARCH=""` implies PETSC being prefix-installed (with installation finished) with `PETSC_DIR=$prefix`.

> PERMON does not support prefix - there is no configure.

Yes, I realized that `prefix-installed PETSc => PETSC_ARCH="" => PERMON's lib directory is $PERMON_DIR/lib`. I think it's ok for now.

jkruzik: It is not OK. By empty I mean it is not set. Setting `PETSC_ARCH=""` does not make it empty, you need to `unset PETSC_ARCH`

I now remembered, that I spend quite a while making PERMON work with prefix installs of PETSc https://github.com/permon/permon/commit/bb9dcb1525db64a4acfbb6ce3c40b2785de3703c

In any case, this should be tested with all the 3 configs I mentioned above.

haplav: > It is not OK.

I meant the current PERMON behavior that it does not really support prefix install and installs into its root directory instead.

> By empty I mean it is not set. Setting PETSC_ARCH="" does not make it empty, you need to unset PETSC_ARCH

OK, with PETSC_ARCH being unset, configure just assumes some default arch name. In my case it's `arch-linux2-c-debug`. It does not install everything into PETSC_DIR directly! Anyway, for us it's not a separate option because we then need to set this as `PETSC_ARCH`. So for us only your options 1 and 3 are relevant.

jkruzik: > OK, with PETSC_ARCH being unset, configure just assumes some default arch name. In my case it's `arch-linux2-c-debug`. It does not install everything into PETSC_DIR directly! Anyway, for us it's not a separate option because we then need to set this as `PETSC_ARCH`. So for us only your options 1 and 3 are relevant.

OK.

Did you test the removal of the code above with a prefix PETSc install?

haplav: I now pushed 2f2b027 with the PETSC_ARCH-based decision and it's working for both variants (prefix/no prefix).

gmakefile.test
```diff
@@ -1,6 +1,7 @@
 # -*- mode: makefile-gmake -*-
 
-CONFIGDIR := config
+# Account for PETSc prefix installation (empty/unset PETSC_ARCH)
+PETSCCONFIGDIR = $(if $(PETSC_ARCH), $(PETSC_DIR)/config, $(PETSC_DIR)/share/petsc/examples/config)
```
jkruzik:
```suggestion
PETSCCONFIGDIR ?= $(if $(PETSC_ARCH), $(PETSC_DIR)/config, $(PETSC_DIR)/share/petsc/examples/config)
```
it is already set when gmakefile includes gmakefile.test.

For consistency, this change should also be in gmakefile.

haplav: > For consistency, this change should also be in gmakefile.

Is the change in ef39a54 what you mean?

Otherwise ok?

haplav:
@jkruzik If you are fine with this otherwise, I will squash it into ~3 commits, force-push and merge.

jkruzik (approved):

jkruzik:
> @jkruzik If you are fine with this otherwise, I will squash it into ~3 commits, force-push and merge.

Looks good to me. You can tidy up the commits and merge.

PR #30, opened 2020-05-06
QPCheckNullSpace replaced w/ MatCheckNullSpace.

haplav:
* MatInvSetNullSpace,QPSetOperatorNullSpace automatically call it in debug mode.
* MatInvSetNullSpace,QPSetOperatorNullSpace support R=NULL.


jkruzik (approved):
This is great. I like that the null space check is done only in debug mode.

haplav:
Tested now with #35 merged in locally.

PR #31, opened 2020-05-07
Create CONTRIBUTING.md

jkruzik:


CONTRIBUTING.md
```diff
@@ -0,0 +1,55 @@
...
+
+Before filing a pull request (PR):
+
+- If your contribution can be logically decomposed into 2 or more separate contributions, submit them in sequence with different branches instead of all at once.
+- Include tests which cover any changes to the source code.
+- Before submitting any PR, run the full test suite - i.e `make alltests TIMEOUT=600` on your machine
+
+### Guidelines for maintainers
+
+- All merges to master shall be with no fast forward - i.e. `--no-ff`
```
haplav: Why? GitHub has `Squash and merge` and `Rebase and merge` which are nice but in contradiction with this rule.

jkruzik: You can always easily find what a feature branch contained. See e.g., https://nvie.com/posts/a-successful-git-branching-model/#incorporating-a-finished-feature-on-develop

An exception could be a branch with a single commit.

jkruzik: It is also used in PETSc (we head a similar model for a while which is based on git flow). See https://jedbrown.org/files/20160225-PETSc.pdf

haplav: PETSc uses GitLab. It has different behavior w.r.t. squash&merge - it squashes the branch into a single commit which is then merged with `--no--ff`.

My more general concern is this is not the right place for our internal policies. I think there are a few of us who are maintainers and we can discuss that by other means. So I suggest removing this `Guidelines for maintainers` section completely.

jkruzik: Since we do not have pipelines for automatic testing (I do want to add this), it is a responsibility to anyone who merges to master to ensure that there are any avoidable mistakes. Therefore if you merge something, in most cases, you need to checkout the branch and make sure that everything works correctly. Depending on the changes this means e.g., running (full) test suite or specifically testing other changes. Then it is easy to merge from command line. I do not think we should rely on a specific behaviour of our current hosting (which may change when we start to look at the test pipelines).

As fort the squash and merge, if a PR is a mess the author should squash the commits.

haplav: OK, I agree that now that the CI testing is not in place, the merger should check out the branch if it includes any code changes, and it _can_ be more handy to do the merge right from the command line. However, for documentation changes, for instance, I don't see a reason why not to use squash&merge after the whole discussion.

Furthermore, once we have CI testing, which should be our goal, the game changes completely. Once a pipeline is run successfully on a messy MR, it's better to let GitLab do squash&merge (keeping the "correct" stamp) than to make the authors do the squashing and rerunning the pipeline (because the authors shouldn't be trusted they didn't screw up anything during the rebase). This workflow is very common now in PETSc.

That said, we can discuss how the workflow should be but CONTRIBUTING.md is in principle _not_ the place for the such workflow rules. It can be just said among us or in the wiki. Not this file. This file should be about contributing, not maintaining. It's kind of agreement between us and potential external contributor. The internal rules are volatile whereas CONTRIBUTING should be kept constant as much as possible. It shouldn't be messed up with something like this.

jkruzik: > OK, I agree that now that the CI testing is not in place, the merger should check out the branch if it includes any code changes, and it _can_ be more handy to do the merge right from the command line. However, for documentation changes, for instance, I don't see a reason why not to use squash&merge after the whole discussion.
>
Well doc change is different as I usually do not care about the evolution (fixing typos, etc.) but I do want some history evolution for more complex branches rather than e.g., a single commit changing 2k lines of code. I am OK with an automatic squashing in the case of documentation (as in the actual text not structure changing things of the whole doc) and as I said a single commit feature could be an exception to the no-ff rule. I can't think of any other files than documentation and text files like README where it is reasonable to chuck the whole branch history.

> Furthermore, once we have CI testing, which should be our goal, the game changes completely. Once a pipeline is run successfully on a messy MR, it's better to let GitLab do squash&merge (keeping the "correct" stamp) than to make the authors do the squashing and rerunning the pipeline (because the authors shouldn't be trusted they didn't screw up anything during the rebase). This workflow is very common now in PETSc.
> 

I do not know whether Satish merges with a web browser (I kinda doubt it) but he certainly does not generally squash other people branches. See e.g., https://gitlab.com/petsc/petsc/-/commit/055802ab15e3ff3537bcc057ae4f769faa7bf190. If the branch is really very messy (the one I linked has even a revert but was still merged) the contributor should do a squash because:

- there is going to be a readable history
- if/when there is a bug, see the point above (in an ideal world, being on any commit, it should be possible to compile the code and check if a bug is present)
- it makes the contributor look at the code again, perhaps finding a bug or a better way to do something
- the number of commits is used as an aggregate contributions metric

Disadvantage is that you need to run the tests again.

> That said, we can discuss how the workflow should be but CONTRIBUTING.md is in principle _not_ the place for the such workflow rules. It can be just said among us or in the wiki. Not this file. This file should be about contributing, not maintaining. It's kind of agreement between us and potential external contributor. The internal rules are volatile whereas CONTRIBUTING should be kept constant as much as possible. It shouldn't be messed up with something like this.

I have no problem with having workflow rules elsewhere but they should be codified.

I think of the CONTRIBUTING.md as a set of guidelines about how the contribution should be made and what should be followed by anyone, not only external contributors, when making them. The Disclaimer part is there to ensure the contributions are OK regarding our license, etc.

haplav: > Well doc change is different as I usually do not care about the evolution (fixing typos, etc.) but I do want some history evolution for more complex branches rather than e.g., a single commit changing 2k lines of code. I

> If the branch is really very messy (the one I linked has even a revert but was still merged) the contributor should do a squash

Well, you just wrote what I think as well - sometimes squashing is appropriate, sometimes not. I don't advocate neither. And if substantial squashing is appropriate, then it's indeed better to defer it to the author.

But what I find an elegant use case of automated squashing is when you have a single commit PR and then make several changes just to fix bugs from testing or incorporate reviewers' comments. Often it can be just a few suggestions applied directly from GitLab using the suggestion functionality.

> I do not know whether Satish merges with a web browser (I kinda doubt it) but he certainly does not generally squash other people branches. See e.g., https://gitlab.com/petsc/petsc/-/commit/055802ab15e3ff3537bcc057ae4f769faa7bf190. 

He _mostly_ does merges from the web browser. Why not? The commit above is done automatically by GitLab from the MR title. Usually, he doesn't have to check out the branch at all.

If you look carefully at the MR above, you can see that the merge commit and the record about merging have exactly the same time.

> but he certainly does not generally squash other people branches.

It's not uncommon. Sometimes it is discussed with the author but is within merger's authority. The author can even give a hint and check `Squash commits when merge request is accepted.` checkbox in the MR description.

Look e.g. at my own https://gitlab.com/petsc/petsc/-/merge_requests/2675 carefully. In master, you can just see [one commit with the collapsed changes](https://gitlab.com/petsc/petsc/-/commit/ebd100ed507624cfce4b131695b689282ee5baf9) and [one merge commit](https://gitlab.com/petsc/petsc/-/commit/717c363adb9bec4584ed72196d6ef91f7878c49d), although the MR contained 3 commits before merging. Satish squash&merged online - there is no record in the MR about him pushing any commits.

The latest [merged MR](https://gitlab.com/petsc/petsc/-/merge_requests/2730) is the same.

haplav: OK, if you remove this, I'm otherwise fine with this pull request.
```suggestion
```

haplav: It's actually  a chance for you to at least try the workflow of just clicking `Commit suggestion` above and use squash&merge ;-)

jkruzik: > But what I find an elegant use case of automated squashing is when you have a single commit PR and then make several changes just to fix bugs from testing or incorporate reviewers' comments. Often it can be just a few suggestions applied directly from GitLab using the suggestion functionality.
> 
Agreed, except when it is a code change (with minor exceptions) and we do not have the automated testing.
> He _mostly_ does merges from the web browser. Why not? The commit above is done automatically by GitLab from the MR title. Usually, he doesn't have to check out the branch at all.
> 
As I said I do not know his workflow but if I merge two unrelated branches in succession how do I know that the first one does not break the second one?
> If you look carefully at the MR above, you can see that the merge commit and the record about merging have exactly the same time.
> 
I see 4:14 commit and 4:15 closing of the MR. 

> Look e.g. at my own [gitlab.com/petsc/petsc/-/merge_requests/2675](https://gitlab.com/petsc/petsc/-/merge_requests/2675) carefully. In master, you can just see [one commit with the collapsed changes](https://gitlab.com/petsc/petsc/-/commit/ebd100ed507624cfce4b131695b689282ee5baf9) and [one merge commit](https://gitlab.com/petsc/petsc/-/commit/717c363adb9bec4584ed72196d6ef91f7878c49d), although the MR contained 3 commits before merging. Satish squash&merged online - there is no record in the MR about him pushing any commits.
> 
Again, I see there about a minute discrepancy between the timestamps.

Anyway, I think this is getting a bit offtopic. I do want to have written guidelines for maintainers. We should be consistent with how we approach the maintenance. So the question is where they should be a where we can discuss them?

haplav: > I see 4:14 commit and 4:15 closing of the MR.

I see 12:15am both :-)

Screenshot 2020-05-08 at 10 22 58

Screenshot 2020-05-08 at 10 23 07

haplav: > I do want to have written guidelines for maintainers. We should be consistent with how we approach the maintenance. So the question is where they should be a where we can discuss them?

[Wiki](https://github.com/permon/permon/wiki/PERMON-Wiki). It is versioned and it is in MD. We can port it easily to another host in future if needed.

haplav: Actually there are quite a few more formats supported by GitHub. E.g. RST is now adopted for https://docs.petsc.org/en/latest/ But I honestly don't like that format as much as MD.

haplav (approved):

CONTRIBUTING.md
jkruzik: > I see 12:15am both :-)
> 
I meant 12:14 - it is the actual time of the commit ...2ab - a minute before GitLab automatically closed the MR. I tried it out on one of my private repos on GitLab: I did a merge request, no-ff merge from CMD, waited couple of minutes and then push. The timestamp on the git commit is my local one, and the "mentioned in commit" and "merged" timestamp correspond to the push time.

haplav:
@jkruzik @mpecha This should now be copied also to PermonSVM, I guess?

PR #35, opened 2020-05-13
update src/tutorials/output/jbearing2_3.out

### LINKED ISSUE #19: tutorials-jbearing2_3 test issue
1. I get
   ```
   not ok tutorials-jbearing2_3 # Exceeded timeout limit of 60 s
   ```
   so I added
   ```
     timeoutfactor: 2
   ```
2. But then I get
```
not ok diff-tutorials-jbearing2_3 # Error code: 1
#	8,9c8,9
#	<   last QPSSolve CONVERGED due to CONVERGED_ATOL, KSPReason=3, required 2361 iterations
#	<   all 1 QPSSolves from last QPSReset/QPSResetStatistics have required 2361 iterations
#	---
#	>   last QPSSolve CONVERGED due to CONVERGED_ATOL, KSPReason=3, required 2637 iterations
#	>   all 1 QPSSolves from last QPSReset/QPSResetStatistics have required 2637 iterations
#	18c18
#	<         Gradient steps: 678
#	---
#	>         Gradient steps: 780
#	28,30c28,30
#	<             Number of updates: 2352
#	<             Number of rejects: 8
#	<             Number of resets: 688
#	---
#	>             Number of updates: 2635
#	>             Number of rejects: 1
#	>             Number of resets: 785
#	39c39
#	<                 Number of updates: 2352
#	---
#	>                 Number of updates: 2635
#	41c41
#	<                 Number of resets: 696
#	---
#	>                 Number of resets: 786
#	53c53
#	<         Residual in Function/Gradient:=9.99023e-09
#	---
#	>         Residual in Function/Gradient:=9.92755e-09
#	55,56c55,56
#	<         total number of iterations=2361,                                (max: 2000)
#	<         total number of function/gradient evaluations=5109,            (max: 4000)
#	---
#	>         total number of iterations=2637,                                (max: 2000)
#	>         total number of function/gradient evaluations=5159,            (max: 4000)
#	58c58
#	< norm of difference of results from TAO and QP = 7.624551e-07  
#	---
#	> norm of difference of results from TAO and QP = 3.516641e-07  

# -------------
#   Summary    
# -------------
# FAILED diff-tutorials-jbearing2_3
```

I assume this TAO iteration is not stable enough to yield always the same numbers. But what to do with that? A few possibilities here:
1. limit number of iterations
2. remove `-qps_view_convergence` as the results are compared anyway
3. use some filtering
4. use alt files

haplav:
fix #19


haplav:
@jkruzik I'm not sure whether this output file should be updated or an alt file added. Can it be just out of date? Does it pass for you currently?

jkruzik:
The current master works for me with PETSC maint (a4758970) with both Clang (10.0.0) and GCC (9.3.0) on AMD Zen 1st gen and I believe it works fine on Intel Broadwell and Coffee Lake with Clang (I will check it in the evening).

The outputs (which were tested on the 3 architectures mentioned above) did not really change since v3.10 (4d1982e9ce4aa19a0dfab414e9694f0d6448dcd2), where there was a big performance regression.

As for the timings, I get:
```
# Timing summary (actual test time / total CPU time): 
#   tutorials-jbearing2_3: 17.71 sec / 33.84 sec
#   tutorials-jbearing2_6: 8.95 sec / 17.09 sec
```
Perhaps we can reduce the problem size (say mx=my=30) for tests 3 and 6 to reduce the time required and it may fix the BLMVM instability.

haplav:
@jkruzik 
> Perhaps we can reduce the problem size (say mx=my=30) for tests 3 and 6 to reduce the time required and it may fix the BLMVM instability.

Sounds good, will try.

haplav:
I reduced it by half to 25 - sorry haven't noticed your suggestion of 30.

haplav:
Let me also push a suggestion addressing my previous concern https://github.com/permon/permon/issues/19#issuecomment-620623637.

haplav:
OK, @jkruzik, tested on my end and ready for your review!

src/tutorials/jbearing2.c
```diff
@@ -540,12 +542,18 @@ PetscErrorCode CallPermonAndCompareResults(Tao tao, void *ctx)
   ierr = QPGetSolutionVector(qp, &x_qp);CHKERRQ(ierr);
   ierr = TaoGetSolutionVector(tao,&x_tao);CHKERRQ(ierr);
   
-  /*Different of results from Snes and QP */
+  /* Different of results from Snes and QP */
+  ierr = VecNorm(user->B, NORM_2, &rhs_norm);CHKERRQ(ierr);
+  tao_diff_tol = 1e2 * rtol / rhs_norm;
```
jkruzik:
```suggestion
  tao_diff_tol = 1e1 * PetscMax(rtol*rhs_norm,atol); /* 10 times the tolerance for the default convergence test */
```
Throws an error when the difference is more than 10* the stopping tolerance and mentions to users what is the default tolerance.

src/tutorials/jbearing2.c
```diff
@@ -540,12 +542,18 @@ PetscErrorCode CallPermonAndCompareResults(Tao tao, void *ctx)
...
+  ierr = PetscOptionsGetReal(NULL, NULL, "-tao_diff_tol", &tao_diff_tol, NULL);CHKERRQ(ierr);
+  ierr = PetscOptionsGetBool(NULL, NULL, "-tao_diff_print", &print_tao_diff, NULL);CHKERRQ(ierr);
   ierr = VecCopy(x_tao, x_diff);CHKERRQ(ierr);
   ierr = VecAXPY(x_diff, -1.0, x_qp);CHKERRQ(ierr);
   ierr = VecNorm(x_diff, NORM_2, &x_diff_norm);CHKERRQ(ierr);
-  ierr = PetscPrintf(PETSC_COMM_WORLD,"norm of difference of results from TAO and QP = %e  \n",x_diff_norm);CHKERRQ(ierr);
-  
+  if (x_diff_norm > tao_diff_tol) SETERRQ2(PetscObjectComm((PetscObject)qps), PETSC_ERR_PLIB, "PERMON and TAO yield different results!\nNorm of difference of results from TAO and QP = %e > %e = tolerance",x_diff_norm,tao_diff_tol);
+  if (print_tao_diff) {
```
jkruzik: I think we could keep the print by default (we are already printing the tolerances). Moreover, this is an example so getting some feedback is good.

I would suggest getting rid of the option altogether.

src/tutorials/jbearing2.c
```diff
@@ -540,12 +542,18 @@ PetscErrorCode CallPermonAndCompareResults(Tao tao, void *ctx)
   ierr = QPGetSolutionVector(qp, &x_qp);CHKERRQ(ierr);
   ierr = TaoGetSolutionVector(tao,&x_tao);CHKERRQ(ierr);
   
-  /*Different of results from Snes and QP */
+  /* Different of results from Snes and QP */
```
jkruzik:
```suggestion
  /* Difference of results from TAO and QP */
```

src/tutorials/jbearing2.c
haplav: Fair enough, I will remove it.

src/tutorials/jbearing2.c
haplav: Oh, good point that rtol must be _multiplied_ with rhs_norm.

haplav: OK, I will apply it and make sure all tests still pass.

haplav: @jkruzik Unfortunately,
```
mpirun -n 2 ./jbearing2 -mx 100 -my 100
---- Journal Bearing Problem SHB-----
mx: 100,  my: 100,  ecc: 0.1 

Value of Tao are gatol = 1.000000e-08, grtol =  1.000000e-08, gttol = 0.000000e+00  
Norm of difference of results from TAO and QP = 4.807057e-06 > 1.000000e-07 = tolerance
[0]PETSC ERROR: --------------------- Error Message --------------------------------------------------------------
[0]PETSC ERROR: Petsc has generated inconsistent data
[0]PETSC ERROR: PERMON and TAO yield different results!
...
```
so I would like to return to multiplying by 1e2, although such big problem is not tested now.

haplav: Actually, as you can see, even 1e-6 won't be big enough here to pass :-( Suggestions?

jkruzik: I think we do not have the same stopping criteria:
```
Value of Tao are gatol = 1.000000e-08, grtol =  1.000000e-08, gttol = 0.000000e+00
```
According to [TaoSetTolerances()](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Tao/TaoSetTolerances.html), grtol is relative to f(x) while, assuming x_0 = 0, gttol == rtol in QPS.

The same issue is in: https://github.com/permon/permon/blob/e2bcc13e1edf12299bb0d05d2fe67cd3074ee696/src/qps/impls/tao/qpstao.c#L188

Maybe you can try running with `-tao_converged_reason` to see what we are stopping on (I think it is grtol).

src/tutorials/jbearing2.c
```diff
@@ -493,9 +493,10 @@ PetscErrorCode CallPermonAndCompareResults(Tao tao, void *ctx)
   PetscReal      rtol,atol,dtol;
   PetscInt       maxit;
   PetscReal      gatol, grtol, gttol;
+  PetscReal      rhs_norm;
+  PetscReal      tao_diff_tol = 1e2*PETSC_SQRT_MACHINE_EPSILON;
```
jkruzik, line 502: Remove the assignment - the default is set later. Maybe it can be grouped with the PetscReal above.

src/tutorials/jbearing2.c
```diff
@@ -583,6 +587,5 @@ PetscErrorCode CallPermonAndCompareResults(Tao tao, void *ctx)
     test:
       suffix: 6
       nsize: 3
-      args: -mx 50 -my 50
+      args: -mx 30 -my 30
 TEST*/
-
```
jkruzik, line 588: Please keep the empty line at the end of the code because some compilers (CRAY) show a warning when the file is not POSIX compliant.

src/tutorials/jbearing2.c
haplav, line 497: @jkruzik Please commit suggestion if ok:
```suggestion
  PetscReal      rhs_norm, tao_diff_tol;
```

haplav, line 502: @jkruzik Please commit suggestion if ok:
```suggestion
  PetscReal      rhs_norm, tao_diff_tol;
```
(notice GitHub support multi-line suggestions)

src/tutorials/jbearing2.c
haplav: * I'm not aware I would do this explicitly, maybe GitHub or some other tool did that automatically.
* I don't think you're right here. The source file should end with `\n`. Which Vim does by default, I think. No need for blank line - that would actually be `\n\n`.
https://unix.stackexchange.com/a/18789
* PETSc does not have blank line at EOF everywhere: 
  - https://gitlab.com/petsc/petsc/-/blob/master/src/dm/impls/plex/tests/ex1.c
  - https://gitlab.com/petsc/petsc/-/blob/master/src/ksp/ksp/interface/itcreate.c
* Git actually deems it a whitespace error:
```
   $ git diff
diff --git a/src/tutorials/jbearing2.c b/src/tutorials/jbearing2.c
index 23244b3..e4adbac 100644
--- a/src/tutorials/jbearing2.c
+++ b/src/tutorials/jbearing2.c
@@ -589,3 +589,4 @@ PetscErrorCode CallPermonAndCompareResults(Tao tao, void *ctx)
       nsize: 3
       args: -mx 30 -my 30
 TEST*/
+
$ # + above is printed with red background
$ git diff --check
src/tutorials/jbearing2.c:592: new blank line at EOF.
```

src/tutorials/jbearing2.c
haplav: Hm, maybe we could have the TAO stopping criterion (relative to f(x)) as an alternative stopping criterion in PERMON? It would be cool if could get exact 0.0 for TAO and QPSTAO with the same settings.

src/tutorials/jbearing2.c
haplav: So I think the blank line before EOF should actually be considered erroneous rather than required :-)

jkruzik: > `# + above is printed with red background`
> `new blank line at EOF`

On the newest git, I get this warning only when there is 3 or more successive \n. I.e., having `text\n\nEOF` is considered OK.

Anyway, I agree in this case there would be a double \n, which is unnecessary (but not erroneous). You can keep it as it is.

Both git diff and GitHub web interface highlight missing newlines sufficiently so that there is no need for double \n. (`\ No newline at end of file` in git diff and see ex1.c for GitHub example in https://github.com/permon/permonsvm/pull/1/commits/78ff1ad23b9b6db256da322c02bb2118f2f4119b).

src/tutorials/jbearing2.c
jkruzik: Well, if we fix qpstao (say, ```TaoSetTolerances( tao, qps->atol, PETSC_MACHINE_EPSILON, qps->rtol)```) we should get diff very close to 0.

We could add something like ConvergenceTestTao() that would be the same as default ConvergenceTest in TAO.

Also, we could make an example with a custom convergence test with g(x)/f(x) stopping criterion. Wink, wink PR #40.

haplav: > We could add something like ConvergenceTestTao() that would be the same as default ConvergenceTest in TAO.

Yes, it would be only valid for QPSTAO and would make use of all TAO tests because there are currently more of them than in PERMON.

This brings me to the main problem here: TAO is actually diverging for `-mx` and `-my` over ~30 due to `DIVERGED_MAXFCN`. It seems BLMVM has serious convergence problems with those bigger problems (not our problem, though).

So I suggest adding just checks that TAO and QPS both converge (otherwise it makes no sense to compare them) - will push shortly.

> Also, we could make an example with a custom convergence test with g(x)/f(x) stopping criterion. Wink, wink PR #40.

Yes, I actually have it already. I was even thinking about adding it to the library rather than just example. But the problem is we would have to make sure in all QPS implementation that the objective value is computed in every iteration, or would have an additional MatMult in every stopping criterion evaluation. I don't think we want either now.

So should I add it to this example for now? Within this PR?

haplav:
(rebase onto master)

src/tutorials/jbearing2.c
haplav: > So should I add it to this example for now? Within this PR?

If so, it would be better to get #40 merged first.

haplav:
OK, @jkruzik, this seems to work as it is now. I suggest merging this (after squashing) and do the things of from https://github.com/permon/permon/pull/35#discussion_r426733599 separately once #40 is merged.

src/tutorials/jbearing2.c
jkruzik: > > We could add something like ConvergenceTestTao() that would be the same as default ConvergenceTest in TAO.
> 
> Yes, it would be only valid for QPSTAO and would make use of all TAO tests because there are currently more of them than in PERMON.
> 
I could only find the TaoConvergedDefault. More importantly, I would prefer QPSTAO to use by default our convergence test that is compatible with other solvers in PERMON (i.e., `TaoSetTolerances( tao, qps->atol, PETSC_MACHINE_EPSILON, qps->rtol)` where we need to check if TAO takes x_0 into account). In this way, we can freely swap solvers while having consistent stopping criteria.
> This brings me to the main problem here: TAO is actually diverging for `-mx` and `-my` over ~30 due to `DIVERGED_MAXFCN`. It seems BLMVM has serious convergence problems with those bigger problems (not our problem, though).
> 
Let us switch to TRON, both MPGP and TRON are stopping on ATOL=1e-8 and the difference between the solution is at worst 1e-8.
> So I suggest adding just checks that TAO and QPS both converge (otherwise it makes no sense to compare them) - will push shortly.
> 
Agreed.
> > Also, we could make an example with a custom convergence test with g(x)/f(x) stopping criterion. Wink, wink PR #40.
> 
> Yes, I actually have it already. I was even thinking about adding it to the library rather than just example. But the problem is we would have to make sure in all QPS implementation that the objective value is computed in every iteration, or would have an additional MatMult in every stopping criterion evaluation. I don't think we want either now.
> 
If an algorithm has computed a gradient (and it is saved into qps/ convergence ctx) we could often evaluate cost function cheaply using QPComputeObjectiveFromGradient().
> So should I add it to this example for now? Within this PR?

I think it would be better to do an example (a separate one, not in jbearing2) in #40, to show the use of QPSGetResidualNorm() and QPSGetIterationNumber() (which are almost universally needed due to removals in #40).

I am not sure how useful is a stopping criterion relative to f(x), because you can start with a large f(x) value (e.g. 1e+8) while the value near the solution can be close to zero but equally it could be very large e.g., -1e+8).

haplav: > I could only find the TaoConvergedDefault. More importantly, I would prefer QPSTAO to use by default our convergence test that is compatible with other solvers in PERMON (i.e., TaoSetTolerances( tao, qps->atol, PETSC_MACHINE_EPSILON, qps->rtol) where we need to check if TAO takes x_0 into account). In this way, we can freely swap solvers while having consistent stopping criteria.

Definitely! I just meant that as an alternative, non-default.

haplav: > If an algorithm has computed a gradient (and it is saved into qps/ convergence ctx) we could often evaluate cost function cheaply using QPComputeObjectiveFromGradient().

True.

haplav: > I think it would be better to do an example (a separate one, not in jbearing2) in #40, to show the use of QPSGetResidualNorm() and QPSGetIterationNumber() (which are almost universally needed due to removals in #40).

OK, but please let's keep #40 concise and do that in a separate MR.

> I am not sure how useful is a stopping criterion relative to f(x), because you can start with a large f(x) value (e.g. 1e+8) while the value near the solution can be close to zero but equally it could be very large e.g., -1e+8).

At least it's standard in TAO. But it makes some sense. If your objective is very large, you are not interested in relatively small improvements of the residual, and vice versa. Maybe I would just add it as alternative convergence test which interprets rtol in a different (TAO) way rather than the default (KSP) way.

haplav: > Let us switch to TRON, both MPGP and TRON are stopping on ATOL=1e-8 and the difference between the solution is at worst 1e-8.

OK, but I could still keep rtol=1e-6 here, no? The original 1e-8 sounds a bit like overkill to me.

jkruzik: 1e-6 is fine, but it should be set as `tao_gttol` while e.g., `tao_grtol=1e-50` (also there is no need to set `tao_gatol` to the default)

haplav: Gotcha, KSP/PERMON rtol is indeed like TAO gttol. Kind of confusing, though.

haplav: Moving the leftovers to Issue #43.

haplav:
@jkruzik I think we are getting there 😅 If you approve, I will rebase, collapse, re-test, merge.

haplav:
(pure squash)

haplav:
(pure rebase)

haplav:
Re-tested.

haplav:
@jkruzik Oops, sorry, I think I confused you approval for #38 :-( I will be more careful next time.

PR #38, opened 2020-05-15
QPSSetFromOptions: process -qps_monitor_cancel

### LINKED ISSUE #37: -qps_monitor_cancel missing in QPSSetFromOptions()
Follow up https://github.com/permon/permon/issues/36#issuecomment-629096815

### LINKED ISSUE #42: use PetscFunctionBeginHot in convergence tests, monitors and other looped functions
https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Sys/PetscFunctionBeginHot.html

haplav:
* QPSSetFromOptions: process `-qps_monitor_cancel`
* Remove clutter (mainly docs) in `QPSMonitorSet()` and friends.
* Fix #37.
* Can be tested with
   ```
   $PERMON_DIR/src/tutorials/jbearing2 -qps_type mpgp -qps_monitor_cancel
   ```

### DIFF
--- a/src/qps/interface/qps.c
+++ b/src/qps/interface/qps.c
@@ -889,7 +889,10 @@ PetscErrorCode QPSSetFromOptions(QPS qps)
   TRY( QPSSetTolerances(qps,rtol,atol,dtol,maxit) );
   TRY( PetscOptionsBool("-qps_auto_post_solve","QPSSolve automatically triggers PostSolve","QPSSetAutoPostSolve",qps->autoPostSolve,&qps->autoPostSolve,NULL) );
   flg = PETSC_FALSE;
-  TRY( PetscOptionsBool("-qps_monitor","Switches QPS monitor","QPSMonitorSet",flg,&flg,NULL) );
+  TRY( PetscOptionsBool("-qps_monitor_cancel","Turn off all QPS monitors","QPSMonitorCancel",flg,&flg,NULL) );
+  if (flg) TRY( QPSMonitorCancel(qps) );
+  flg = PETSC_FALSE;
+  TRY( PetscOptionsBool("-qps_monitor","Turn on default QPS monitor","QPSMonitorSet",flg,&flg,NULL) );
   if (flg) TRY( QPSMonitorSet(qps,QPSMonitorDefault,NULL,NULL) );
   /* actually checked in setup - this is just here to go into help message */
   TRY( PetscOptionsName("-qps_view","print the QPS parameters at the end of a QPSSolve call","QPSView",&flg) );
@@ -1132,7 +1135,7 @@ PetscErrorCode QPSSolutionVecStateChanged(QPS qps,PetscBool *flg)
 -  rnorm - relative norm of the residual
 
    Notes:
-   This routine is called by the KSP implementations.
+   This routine is called by the QPS implementations.
    It does not typically need to be called by the user.
 
    Level: developer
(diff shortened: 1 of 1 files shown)


haplav:
(pure rebase onto master)

src/qps/interface/qps.c
```diff
@@ -889,8 +889,11 @@ PetscErrorCode QPSSetFromOptions(QPS qps)
   TRY( QPSSetTolerances(qps,rtol,atol,dtol,maxit) );
   TRY( PetscOptionsBool("-qps_auto_post_solve","QPSSolve automatically triggers PostSolve","QPSSetAutoPostSolve",qps->autoPostSolve,&qps->autoPostSolve,NULL) );
   flg = PETSC_FALSE;
-  TRY( PetscOptionsBool("-qps_monitor","Switches QPS monitor","QPSMonitorSet",flg,&flg,NULL) );
+  TRY( PetscOptionsBool("-qps_monitor","Turn on default QPS monitor","QPSMonitorSet",flg,&flg,NULL) );
   if (flg) TRY( QPSMonitorSet(qps,QPSMonitorDefault,NULL,NULL) );
+  flg = PETSC_FALSE;
+  TRY( PetscOptionsBool("-qps_monitor_cancel","Turn off all QPS monitors","QPSMonitorCancel",flg,&flg,NULL) );
+  if (flg) TRY( QPSMonitorCancel(qps) );
```
jkruzik:
```suggestion
  TRY( PetscOptionsBool("-qps_monitor_cancel","Turn off all QPS monitors","QPSMonitorCancel",flg,&flg,NULL) );
  if (flg) TRY( QPSMonitorCancel(qps) );
  flg = PETSC_FALSE;
  TRY( PetscOptionsBool("-qps_monitor","Turn on default QPS monitor","QPSMonitorSet",flg,&flg,NULL) );
  if (flg) TRY( QPSMonitorSet(qps,QPSMonitorDefault,NULL,NULL) );
```
Cancel hardwired monitors but allow a user to set monitor from options. (E.g., I set a custom monitor routine in the code and I do want to turn that off and run with the default one.)

src/qps/interface/qps.c
```diff
@@ -1160,75 +1163,46 @@ PetscErrorCode QPSMonitor(QPS qps, PetscInt it, PetscReal rnorm)
...
+-  monitordestroy - [optional] routine that frees monitor context (may be NULL)
 
    Calling Sequence of monitor:
-$     monitor (QPS qps, int it, PetscReal rnorm, void *mctx)
+$     monitor (QPS qps, PetscInt it, PetscReal rnorm, void *mctx)
 
 +  qps - iterative context obtained from QPSCreate()
 .  it - iteration number
-.  rnorm - (estimated) 2-norm of (preconditioned) residual
+.  rnorm - estimated 2-norm of the residual
```
jkruzik: When it is estimated?

Strictly speaking, we usually do not deal with a plain residual but I do not have a better wording.

src/qps/interface/qps.c
```diff
@@ -1160,75 +1163,46 @@ PetscErrorCode QPSMonitor(QPS qps, PetscInt it, PetscReal rnorm)
...
-                             uses QPSMonitorLGResidualNormCreate()
-.  -qps_monitor_lg_true_residualnorm   - sets line graph monitor,
-                             uses KSPMonitorLGResidualNormCreate()
-.  -qps_monitor_singular_value    - sets QPSMonitorSingularValue()
--  -qps_monitor_cancel - cancels all monitors that have
-         been hardwired into a code by
-         calls to QPSMonitorSet(), but
-         does not cancel those set via
-         the options database.
+-  -qps_monitor_cancel - Cancels all monitors set from options or with QPSMonitorSet().
```
jkruzik:
```suggestion
-  -qps_monitor_cancel - cancels all monitors that have
         been hardwired into a code by
         calls to QPSMonitorSet(), but
         does not cancel those set via
         the options database.
```
see above

src/qps/interface/qps.c
```diff
@@ -1160,75 +1163,46 @@ PetscErrorCode QPSMonitor(QPS qps, PetscInt it, PetscReal rnorm)
...
-   /*
-   for (i=0; i<qps->numbermonitors;i++) {
-     if (monitor == qps->monitor[i] && monitordestroy == qps->monitordestroy[i] && mctx == qps->monitorcontext[i]) {
-       if (monitordestroy) {
-         (*monitordestroy)(&mctx);
-       }
-       return(0);
-     }
-   }
-   */
```
jkruzik, line 1226: Not sure why this was commented out, but (dropping the second if) this would ensure that two exactly same monitors are not set. I think that is reasonable.

src/qps/interface/qps.c
```diff
@@ -1244,15 +1218,11 @@ PetscErrorCode  QPSMonitorSet(QPS qps,PetscErrorCode (*monitor)(QPS,PetscInt,Pet
 .  qps - iterative context obtained from QPSCreate()
 
    Options Database Key:
-.  -qps_monitor_cancel - Cancels all monitors that have
-     been hardwired into a code by calls to QPSMonitorSet(),
-     but does not cancel those set via the options database.
+.  -qps_monitor_cancel - Cancels all monitors set from options or with QPSMonitorSet().
```
jkruzik: see above

jkruzik (changes requested):

src/qps/interface/qps.c
haplav: OK, can we just reduce it to two lines? This looks terrible :-D
```suggestion
-  -qps_monitor_cancel - cancels all monitors that have been hardwired into a code by
    calls to QPSMonitorSet(), but does not cancel those set via the options database.
```

src/qps/interface/qps.c
haplav: Comparing the destroy function and context looks very weird to me. They don't define the monitor, and it's unlikely exactly the same context pointer would be passed. Also, if the user sets the same monitor twice, it is his wish :-) What about just making sure that `-qps_monitor` sets the monitor just once?

src/qps/interface/qps.c
haplav:
```suggestion
.  -qps_monitor_cancel - cancels all monitors that have been hardwired into a code by
    calls to QPSMonitorSet(), but does not cancel those set via the options database.
```

src/qps/interface/qps.c
haplav: Maybe approximate is better word?

I would say we use the word residual quite consistently as a vector that is evaluated for convergence.

```suggestion
.  rnorm - approximate 2-norm of the residual (more general, a vector whose norm determines convergence, such as a projected gradient for QPSMPGP)
```

haplav: But the latter probably appears multiple times across PERMON.

jkruzik: My point was, that when we compute rnorm it is probably never estimated or approximated.

src/qps/interface/qps.c
jkruzik: I looked into it and if you check SetMonitor routines (e.g., [TSMonitorSet()](https://www.mcs.anl.gov/petsc/petsc-current/src/ts/interface/ts.c.html#TSMonitorSet)) a similar check is done using [PetscMonitorCompare()](https://www.mcs.anl.gov/petsc/petsc-current/src/sys/classes/viewer/interface/viewers.c.html#PetscMonitorCompare).
The docs say that `Checks if two monitors are identical; if they are then it destroys the new one`, but they are identical iff ctx == newctx, so that destroy on newctx also destroys ctx or am I missing something?

I think it does not make sense to call the same monitors (having the same ctx) more than once. On the other hand, it makes sense to set multiple identical monitor routines with different contexts.

haplav: > The docs say that Checks if two monitors are identical; if they are then it destroys the new one, but they are identical iff ctx == newctx, so that destroy on newctx also destroys ctx or am I missing something?

Yeah, I think it doesn't make any sense. Could be filed as a bug actually...

I don't think we should necessarily duplicate the PETSc behavior which is IMHO odd.

> I think it does not make sense to call the same monitors (having the same ctx) more than once. On the other hand, it makes sense to set multiple identical monitor routines with different contexts.

What is the difference between two monitors with the same monitor function and different context? It's still the same monitor and should print exactly the same. The context is just a helper thing.

---
Aside, I think PETSc overuses this model (callback + context + destroy). Sometimes it would be more elegant and less error-prone to have a PetscObject which has its own methods. An extreme case are some functions related to discretization, see https://gitlab.com/petsc/petsc/-/merge_requests/2793#note_343818944.

jkruzik: > > I think it does not make sense to call the same monitors (having the same ctx) more than once. On the other hand, it makes sense to set multiple identical monitor routines with different contexts.
> 
> What is the difference between two monitors with the same monitor function and different context? It's still the same monitor and should print exactly the same. The context is just a helper thing.
> 
You could have a variable in ctx by which the monitor decides what and how it is printed. E.g., you could have a monitor on a subset of a problem (e.g., using a specific submatrix) and then it also makes sense to monitor multiple subsets with the same monitor (but different ctx).

I think this is exactly the reason why there are contexts for monitors.

src/qps/interface/qps.c
haplav: Well, everything is approximate on computers ;-)
```suggestion
.  rnorm - 2-norm of the residual (more general, a vector whose norm determines convergence, such as a projected gradient for QPSMPGP)
```

src/qps/interface/qps.c
haplav: OK

haplav: OK, actually we could just use PetscMonitorCompare(), and create a petsc merge request fixing that. I think if the contexts are identical, it should just return true but not call destroy.

I'm also not persuaded it makes sense to compare the destroy functions.

jkruzik: I think we can leave the `for` loop without the `if (monitordestroy)` for now. Then this can be merged.

haplav: Keep it commented out or uncomment?

jkruzik: Uncomment, without the wrong if.

haplav: OK

haplav:
@jkruzik looks like I have incorporated all your comments. If you approve, I will rebase, collapse, re-test, merge.

Note that I noticed multiple whitespace issues in the monitor-related functions but I will fix it in a separate PR not to mess this one up completely.

jkruzik:
> Note that I noticed multiple whitespace issues in the monitor-related functions but I will fix it in a separate PR not to mess this one up completely.

What kind of issues? If these are in code style, I would prefer fixing them across the whole code base, see https://github.com/permon/permonsvm/issues/3. I do have some scripts for that already, but I need to set aside at least a few days for the fix so that everything can be checked.

jkruzik (approved):

haplav:
(pure squash)

haplav:
(pure rebase, re-tested)

haplav:
> > Note that I noticed multiple whitespace issues in the monitor-related functions but I will fix it in a separate PR not to mess this one up completely.
> 
> What kind of issues? If these are in code style, I would prefer fixing them across the whole code base, see [permon/permonsvm#3](https://github.com/permon/permonsvm/issues/3). I do have some scripts for that already, but I need to set aside at least a few days for the fix so that everything can be checked.

See MR #45. Just some initial spaces. I think it arose when copying codes from PETSc somehow. I think we don't need to wait as it is limited to a small part of one source file.

PR #40, opened 2020-05-18
QPS: Reduce number of arguments of convergence tests.

haplav:
Solved QP, iteration and residual norm are available from the QPS struct.

### DIFF
--- a/src/qps/impls/smalxe/smalxe.c
+++ b/src/qps/impls/smalxe/smalxe.c
@@ -616,9 +616,9 @@ PETSC_STATIC_INLINE PetscErrorCode QPSConverged_Inner_SMALXE_Monitor_Inner(QPS q
 
 #undef __FUNCT__
 #define __FUNCT__ "QPSConverged_Inner_SMALXE"
-PetscErrorCode QPSConverged_Inner_SMALXE(QPS qps_inner,KSPConvergedReason *reason,void *ctx)
+PetscErrorCode QPSConverged_Inner_SMALXE(QPS qps_inner,KSPConvergedReason *reason)
 {
-  QPSConvergedCtx_Inner_SMALXE *cctx = (QPSConvergedCtx_Inner_SMALXE*) ctx;
+  QPSConvergedCtx_Inner_SMALXE *cctx = (QPSConvergedCtx_Inner_SMALXE*) qps_inner->cnvctx;
   QPS qps_outer = cctx->qps_outer;
   QPS_SMALXE *smalxe = (QPS_SMALXE*)qps_outer->data;
   QP qp_inner = qps_inner->solQP;
@@ -655,7 +655,7 @@ PetscErrorCode QPSConverged_Inner_SMALXE(QPS qps_inner,KSPConvergedReason *reaso
   }
 
   
-  TRY( (*qps_outer->convergencetest)(qps_outer,&qps_outer->reason,qps_outer->cnvctx) );
+  TRY( (*qps_outer->convergencetest)(qps_outer,&qps_outer->reason) );
 
   if (qps_outer->reason) {
     if (qps_outer->reason > 0) {
--- a/src/qps/impls/smalxe/smalxe.c
+++ b/src/qps/impls/smalxe/smalxe.c
@@ -628,10 +628,7 @@ PetscErrorCode QPSConverged_Inner_SMALXE(QPS qps_inner,KSPConvergedReason *reaso
   MPI_Comm comm;
   
   PetscFunctionBegin;
-  FLLOP_ASSERT(qp_inner->parent == cctx->qp_outer,"qp_inner->parent == qp_outer");
-  FLLOP_ASSERT(qp_inner->x == cctx->qp_outer->x, "qp_inner->x == qp_outer->x");
   TRY( PetscObjectGetComm((PetscObject)qps_inner,&comm) );
-  if (!qps_inner->setupcalled) FLLOP_SETERRQ(comm, PETSC_ERR_ARG_WRONGSTATE, "Inner QPSSetUp() not yet called");
   *reason = KSP_CONVERGED_ITERATING;
 
   TRY( smalxe->updateNormBu(qps_outer,u,&smalxe->normBu,&smalxe->enorm) );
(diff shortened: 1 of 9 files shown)


src/qps/impls/smalxe/smalxe.c
```diff
@@ -616,19 +616,22 @@ PETSC_STATIC_INLINE PetscErrorCode QPSConverged_Inner_SMALXE_Monitor_Inner(QPS q
...
+  PetscReal gnorm = qps_inner->rnorm;
   MPI_Comm comm;
   
   PetscFunctionBegin;
-  FLLOP_ASSERT(qp_inner->parent == qp_outer,"qp_inner->parent == qp_outer");
-  FLLOP_ASSERT(qp_inner->x == qp_outer->x, "qp_inner->x == qp_outer->x");
+  FLLOP_ASSERT(qp_inner->parent == cctx->qp_outer,"qp_inner->parent == qp_outer");
+  FLLOP_ASSERT(qp_inner->x == cctx->qp_outer->x, "qp_inner->x == qp_outer->x");
   TRY( PetscObjectGetComm((PetscObject)qps_inner,&comm) );
+  if (!qps_inner->setupcalled) FLLOP_SETERRQ(comm, PETSC_ERR_ARG_WRONGSTATE, "Inner QPSSetUp() not yet called");
```
jkruzik: I do not think that this is neccessary.

src/qps/interface/qps.c
```diff
@@ -683,14 +681,19 @@ PetscErrorCode QPSGetConvergenceContext(QPS qps,void **ctx)
...
+PetscErrorCode QPSConvergedDefault(QPS qps,KSPConvergedReason *reason,void *ctx)
 {
   QPSConvergedDefaultCtx *cctx = (QPSConvergedDefaultCtx*) ctx;
+  QP qp = qps->solQP;
+  PetscInt i = qps->iteration;
+  PetscReal rnorm = qps->rnorm;
 
   PetscFunctionBegin;
   PetscValidHeaderSpecific(qps,QPS_CLASSID,1);
+  PetscValidHeaderSpecific(qp,QP_CLASSID,2);
```
jkruzik:
```suggestion
```

src/qps/interface/qps.c
```diff
@@ -683,14 +681,19 @@ PetscErrorCode QPSGetConvergenceContext(QPS qps,void **ctx)
...
   *reason = KSP_CONVERGED_ITERATING;
 
+  if (!qps->setupcalled) FLLOP_SETERRQ(((PetscObject) qps)->comm, PETSC_ERR_ARG_WRONGSTATE, "QPSSetUp() not yet called");
```
jkruzik: Again, is this necessary?

src/qps/impls/smalxe/smalxe.c
haplav: but does not hurt neither - kind of assertion check

src/qps/interface/qps.c
haplav: oops

src/qps/interface/qps.c
haplav: Here it's even more important. We need to be sure e.g. that `solQP` is in place.

src/qps/impls/smalxe/smalxe.c
jkruzik: My point is that we know that QPSSetUp(qps_inner) was called because at this point QPSSetUp(qps_outter) was called and created qps_inner.

src/qps/interface/qps.c
jkruzik: OK. However, I would suggest putting this in QPSConvergedDefaultSetUp(), which also computes norm of solQP->b and then this check will be done only once.

src/qps/interface/qps.c
```diff
@@ -683,14 +681,19 @@ PetscErrorCode QPSGetConvergenceContext(QPS qps,void **ctx)
...
   if (!cctx->setup_called) {
     TRY( QPSConvergedDefaultSetUp(ctx,qps) );
   }
```
jkruzik, line 697: I can't comment on the line bellow, but this `if` should be swapped with the following one.

jkruzik:
Do we need `ctx` as an argument to the convergence test? It is stored in `qps->cnvctx`, and we have set/get functions for it.

src/qps/interface/qps.c
haplav: OK, this makes sense.

haplav:
> Do we need `ctx` as an argument to the convergence test? It is stored in `qps->cnvctx`, and we have set/get functions for it.

True, let's remove this arg as well.

The only advantage of this arg might be that no call to the context getter is needed but source files which implement stopping criterions will usually include private headers so `qps->cnvctx` is directly accessible anyway.

src/qps/interface/qps.c
haplav: yes

src/qps/impls/smalxe/smalxe.c
haplav: ok, I moved all the checks here to Solve and only in debug mode

haplav:
> Do we need ctx as an argument to the convergence test? It is stored in qps->cnvctx, and we have set/get functions for it.

meta-note: better start off a thread next time in such cases

haplav:
@jkruzik anything else? If it's good to go already, I prefer to squash it on my own to decide about the commit message.

jkruzik (approved):
Looks good!

jkruzik:
> > Do we need ctx as an argument to the convergence test? It is stored in qps->cnvctx, and we have set/get functions for it.
> 
> meta-note: better start off a thread next time in such cases

I do not know what you mean. There wasn't any thread related to this.

src/qps/impls/smalxe/smalxe.c
```diff
@@ -616,18 +616,18 @@ PETSC_STATIC_INLINE PetscErrorCode QPSConverged_Inner_SMALXE_Monitor_Inner(QPS q
 
 #undef __FUNCT__
 #define __FUNCT__ "QPSConverged_Inner_SMALXE"
-PetscErrorCode QPSConverged_Inner_SMALXE(QPS qps_inner,QP qp_inner,PetscInt i,PetscReal gnorm,KSPConvergedReason *reason,void *ctx)
+PetscErrorCode QPSConverged_Inner_SMALXE(QPS qps_inner,KSPConvergedReason *reason)
```
haplav, line 619: I meant you could start a new thread e.g. here. I like threads more if applicable because they can be marked as resolved and collapsed.

haplav: @jkruzik

PR #46, opened 2020-05-22
fix infinite afeas in qpc

jkruzik:
e.g., `lb = 0, ub = inf, d_i <= 0` -> afeas should be infinity

Reported-by: Adrian Keith Turner <akt@lanl.gov>


haplav:
Cool, but could we have a reproducing example?

jkruzik:
> Cool, but could we have a reproducing example?

Unless you want to call `QPCFeas()` on the situation above and check that it returned Inf, then no.

haplav:
Does the infinite step make sense? Should it be replaced by some big number instead?

jkruzik:
QPCFeas returns maximal afeas s.t. `x - afeas*d` is feasible. With the situation above the direction goes from the bound so that afeas=Inf. Returning anything smaller than Inf defeats the purpose of QPCFeas. It is up to the algorithm to properly deal with infinite afeas.

haplav:
> It is up to the algorithm to properly deal with infinite afeas.

Yes, I agree. And OK, you made me check the code of [QPSMPGP](https://github.com/permon/permon/blob/master/src/qps/impls/mpgp/mpgp.c#L369) :-) So `if (acg <= afeas)` which is always true for `afeas=Inf`, no further operation is done with `afeas`, so looks good.

haplav (approved):

PR #47, opened 2020-05-28
WIP: move nullspace to mat

haplav:


haplav:
@jkruzik May I ask you for a preliminary review? I will finish things like writing manpages and finishing some of the TODOs once the concept is agreed.

src/mat/impls/inv/matinv.c
```diff
@@ -172,53 +172,21 @@ static PetscErrorCode MatInvComputeNullSpace_Inv(Mat imat)
   }
   TRY( MatAssemblyBegin(R,MAT_FINAL_ASSEMBLY) );
   TRY( MatAssemblyEnd(R,MAT_FINAL_ASSEMBLY) );
+  //TODO mem leak?
+  TRY( MatOrthColumns(R, orthType, orthForm, &R, NULL) );
```
jkruzik: I do not think that orthogonalization should be part of MatComputeNullSpaceMat().

src/mat/impls/inv/matinv.c
```diff
@@ -50,14 +50,14 @@ static PetscErrorCode MatInvGetRegularizationType_Inv(Mat imat,MatRegularization
   PetscFunctionReturn(0);
 }
 
+//TODO MOVE THIS TO permonmatnullspace.c
+//TODO Passing the PC is awkward. We will want to avoid that but it needs deeper refactoring.
```
jkruzik, line 54: We could have MatComputeNullSpaceMat() interface function that calls implementation-specific functions and which defaults to MatComputeNullSpaceMat_Default(Mat,PC,...). MatInv would then have only a wrapper implementation to pass PC to _Default().

src/mat/impls/inv/matinv.c
```diff
@@ -50,14 +50,14 @@ static PetscErrorCode MatInvGetRegularizationType_Inv(Mat imat,MatRegularization
...
 #undef __FUNCT__
-#define __FUNCT__ "MatInvComputeNullSpace_Inv"
+#define __FUNCT__ "MatComputeNullSpaceMat"
 #if defined(PETSC_HAVE_MUMPS)
-static PetscErrorCode MatInvComputeNullSpace_Inv(Mat imat)
+PetscErrorCode MatComputeNullSpaceMat(Mat K, PC pc, MatOrthType orthType, MatOrthForm orthForm, Mat *R_new)
```
jkruzik: We should not return R_New. Instead we should set it to K.

src/mat/impls/inv/matinv.c
```diff
@@ -66,35 +66,34 @@ static PetscErrorCode MatInvComputeNullSpace_Inv(Mat imat)
   MPI_Comm blockComm;
 
   PetscFunctionBeginI;
-  TRY( MatInvGetKSP(imat,&ksp) );
-  TRY( KSPGetPC(ksp,&pc) );
-  TRY( PCGetOperators(pc,&Kl,NULL) );
+  PetscValidHeaderSpecific(K,MAT_CLASSID,1);
+  if (pc) PetscValidHeaderSpecific(pc,PC_CLASSID,2);
+  TRY( PetscObjectTypeCompare((PetscObject)K,MATBLOCKDIAG,&blockdiag) );
+  if (blockdiag) {
```
jkruzik: Having a blockdiag specific implementation calling MatComputeNullSpace_Default() on individual blocks would simplify the implementation.

src/mat/impls/inv/matinv.c
```diff
@@ -107,7 +106,7 @@ static PetscErrorCode MatInvComputeNullSpace_Inv(Mat imat)
       TRY( PCFactorGetMatrix(pc,&F) );
       mumps =(Mat_MUMPS*)F->data;
     } else {
-      TRY( PetscPrintf(PetscObjectComm((PetscObject)imat), "WARNING: Performing extra factorization with MUMPS Cholesky just for nullspace detection. Avoid this by setting MUMPS Cholesky as MATINV solver.\n") );
+      if (pc) TRY( PetscPrintf(PetscObjectComm((PetscObject)K), "WARNING: Performing extra factorization with MUMPS Cholesky just for nullspace detection. Avoid this by setting MUMPS Cholesky as MATINV solver.\n") );
```
jkruzik: Info() ?

src/mat/impls/inv/matinv.c
```diff
@@ -402,16 +370,54 @@ static PetscErrorCode MatInvReset_Inv(Mat imat)
...
+    inv->R_state = -1;
+    PetscFunctionReturn(0);
+  }
+  TRY( PetscObjectGetId((PetscObject)R, &inv->R_id) );
+  TRY( PetscObjectStateGet((PetscObject)R, &inv->R_state) );
+  PetscFunctionReturn(0);
+}
+
+//TODO we can take R from inv-A
+static PetscErrorCode MatInvNullSpaceSame_Inv(Mat imat, Mat R, PetscBool *flg)
```
jkruzik: What are these two functions for? I.e., why does MatInv need to keep these state variables?

src/qp/interface/qptransform.c
```diff
@@ -393,7 +393,7 @@ PetscErrorCode QPTEnforceEqByPenalty(QP qp, PetscReal rho_user, PetscBool rho_di
   /* newA = A + rho*BE'*BE */
   TRY( MatCreatePenalized(qp,rho,&newA) );
   TRY( QPSetOperator(child,newA) );
-  TRY( QPSetOperatorNullSpace(child,qp->R) );
+  //TODO what about nullspace
```
jkruzik: Right A + rho*BE'*BE can have a different nullspace. I would just drop it.

src/mat/impls/inv/matinv.c
haplav: * I believe the nullspace base being ortonormal is a usual requirement. `MatNullSpace` [assumes this](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Mat/MatNullSpaceCreate.html).
* If one doesn't want to perform the orthonormalization,`MAT_ORTH_NONE` can be passed.
* This line taken from `QPTDualize` was actually wrong and would require an additional variable and additional `MatDestroy` since `MatOrthColumns` has nothing like a `MatReuse` argument.

src/mat/impls/inv/matinv.c
haplav: I'm not keen to do that. My ultimate goal is rather to get really rid of the PC argument carefully, including that [terrible play of options database parameters](https://github.com/permon/permon/pull/47/files/a9caa50217df0933e5cd9f3957338052edb5db57#diff-99556492510148a43536b8fdd4e88e24R100). It is misconception posed by current state of affairs in PETSc.

I think keeping this is a reasonable temporary solution and I will create an issue.

src/qp/interface/qptransform.c
haplav: Yes, I think it could be derived but no one really cares at the moment, I guess. Do you think the TODO should be removed?

src/qp/interface/qptransform.c
```diff
@@ -283,7 +283,7 @@ PetscErrorCode QPTEnforceEqByProjector(QP qp)
     TRY( MatCreateProd(comm,3,A_arr,&newA) );
   }
   TRY( QPSetOperator(child,newA) );
-  TRY( QPSetOperatorNullSpace(child,qp->R) );
+  //TODO what about null space
```
haplav: @jkruzik I think here it is easy. `P = I - B'*inv(B*B')*B` so the columns of `B'` span nullspace of `P`, right? In "space language" Im P = Ker B which is orthogonal to Ker P = Im B'.

src/mat/impls/inv/matinv.c
haplav: Like this it is slightly more general and less awkward for some use cases. I also like the clear separation of input and output. I don't see the point of making K an inout argument and changing its properties under the hood.

src/mat/impls/inv/matinv.c
haplav: If R changes, the regularization should be recomputed.

We could also lock further use of SetNullSpaceMat but that wouldn't solve the case when R is the same object but has changed - maybe it's fine?

src/mat/impls/inv/matinv.c
haplav: OK, fixed.

src/mat/impls/inv/matinv.c
haplav: OK, I will try

haplav: Actually, ability to do this speaks for the concept of this PR ;-)

haplav: Done in cf5e155.

src/mat/impls/inv/matinv.c
```diff
@@ -175,16 +160,38 @@ PetscErrorCode MatComputeNullSpaceMat(Mat K, PC pc, MatOrthType orthType, MatOrt
...
+  TRY( PetscObjectQueryFunction((PetscObject)K,"MatComputeNullSpaceMat_C",&func) );
+  if (func) {
+    TRY( (*func)(K,pc,orthType,orthForm,R_new) );
+  } else {
+    TRY( MatComputeNullSpaceMat_Default(K,pc,orthType,orthForm,R_new) );
+  }
   PetscFunctionReturnI(0);
 }
-#endif
+
```
haplav, line 204: oops, removing this line
```suggestion
```

src/mat/impls/inv/matinv.c
```diff
@@ -175,16 +160,38 @@ PetscErrorCode MatComputeNullSpaceMat(Mat K, PC pc, MatOrthType orthType, MatOrt
   TRY( MatOrthColumns(R, orthType, orthForm, R_new, NULL) );
   TRY( PetscObjectSetName((PetscObject)*R_new,"R") );
   TRY( MatDestroy(&R) );
-  PetscFunctionReturnI(0);
+  PetscFunctionReturn(0);
 }
 #else
+static PetscErrorCode MatComputeNullSpaceMat(Mat K, PC pc, MatOrthType orthType, MatOrthForm orthForm, Mat *R_new)
```
haplav: another oops
```suggestion
static PetscErrorCode MatComputeNullSpaceMat_Default(Mat K, PC pc, MatOrthType orthType, MatOrthForm orthForm, Mat *R_new)
```

src/mat/impls/inv/matinv.c
```diff
@@ -175,16 +160,38 @@ PetscErrorCode MatComputeNullSpaceMat(Mat K, PC pc, MatOrthType orthType, MatOrt
...
+{
+  PetscFunctionBegin;
+  FLLOP_SETERRQ(PETSC_COMM_WORLD,PETSC_ERR_SUP,"MUMPS library is currently needed for nullspace computation");
```
haplav:
```suggestion
  FLLOP_SETERRQ(PetscObjectComm((PetscObject)K),PETSC_ERR_SUP,"MUMPS library is currently needed for nullspace computation");
```

src/mat/impls/inv/matinv.c
jkruzik: If R changes the K is still implicitly/explicitly regularized. The actual values in R do not matter as long as it represents a null space basis.

src/qp/interface/qptransform.c
jkruzik: Agreed, but what is Ker P'KP?

src/qp/interface/qptransform.c
jkruzik: I would just make the TODO clearer. Something along the lines of `derive null space of the new operator`?

src/mat/impls/inv/matinv.c
jkruzik: > including that [terrible play of options database parameters](https://github.com/permon/permon/pull/47/files/a9caa50217df0933e5cd9f3957338052edb5db57#diff-99556492510148a43536b8fdd4e88e24R100). It is misconception posed by current state of affairs in PETSc.
> 
What are you referring to?

src/mat/impls/inv/matinv.c
jkruzik: > Like this it is slightly more general and less awkward for some use cases. 

What use case?

I think it is weird to compute the nullspace, which you argue that is the property of a matrix, and not set it to the matrix.

src/mat/impls/inv/matinv.c
jkruzik: OK, it makes sense to choose orthType.

src/mat/impls/inv/matinv.c
```diff
@@ -172,51 +157,39 @@ static PetscErrorCode MatInvComputeNullSpace_Inv(Mat imat)
...
-  *R = inv->R;
-  PetscFunctionReturn(0);
+  PetscErrorCode (*func)(Mat,PC,MatOrthType,MatOrthForm,Mat*)=NULL;
+
+  PetscFunctionBeginI;
+  PetscValidHeaderSpecific(K,MAT_CLASSID,1);
+  if (pc) PetscValidHeaderSpecific(pc,PC_CLASSID,2);
+  PetscValidLogicalCollectiveEnum(K,orthType,3);
+  PetscValidLogicalCollectiveEnum(K,orthForm,4);
+  PetscValidPointer(R_new,5);
```
jkruzik, line 187: We should check if K already contains a null space matrix.

src/mat/impls/inv/matinv.c
haplav: Sorry, probably a wrong link. I meant the use of PetscOptionsInsertString in https://github.com/permon/permon/blob/a9caa50217df0933e5cd9f3957338052edb5db57/src/mat/impls/inv/matinv.c#L103

src/mat/impls/inv/matinv.c
haplav: OK, that makes sense to me. So I will remove this clutter.

src/qp/interface/qptransform.c
haplav:
```suggestion
  //TODO Maybe we can derive the null space of newA?
```

src/qp/interface/qptransform.c
haplav: Since P is a projector and K is multiplied from both sides by P, I believe Ker PKP = Ker P, and the latter is derived above.

src/mat/impls/inv/matinv.c
haplav: I think that a function BlaComputeBla should just compute what is desired and return that without any unseen side effects.

Consider `QPTDualize()`. I think what I have here,
```
TRY( MatComputeNullSpaceMat(K,pc,MAT_ORTH_GS,MAT_ORTH_FORM_EXPLICIT,&R) );
TRY( MatSetNullSpaceMat(K,R) );
TRY( MaDestroy(&R) );
```
is clearer and less error-prone than
```
TRY( MatComputeNullSpaceMat(K,pc,MAT_ORTH_GS,MAT_ORTH_FORM_EXPLICIT) );
//Unclear that R is updated internally.
//Old R needs to be destroyed so the pointer can be dangling now!

TRY( MatGetNullSpaceMat(K,&R) );
```
and requires less documentation.

I think a single clearly defined [mutator](https://en.wikipedia.org/wiki/Mutator_method) function is in general a preferred practice from various reasons. For example, one could hang some callback to the Set function but forget about the other functions altering the property.

src/mat/impls/inv/matinv.c
haplav: Why? What should be the reaction if it contains one already?

Once again, I think that a function BlaComputeBla should just compute what is desired and return that.

src/mat/impls/inv/matinv.c
haplav: Done in 30c209e

src/mat/impls/inv/matinv.c
jkruzik: I am not sure how do you want to remove either PC or the OptionsInsert.

src/qp/interface/qptransform.c
jkruzik: I think Ker PKP is a union of Ker P, Ker KP, and Ker PKP. For instance, Ker KP = Ker K intersect Image P. I.e., Ker PKP must depend on Ker K. Indeed if K is a matrix of all zeros, Ker PKP = R^n = Ker K.

src/mat/impls/inv/matinv.c
jkruzik: There is no point to again compute R if it is already computed. If we decide to return R in this function, we should just do that while incrementing object reference.

src/qp/interface/qptransform.c
haplav: Yes, I agree Ker PKP must depend on Ker K. My [previous reply](https://github.com/permon/permon/pull/47#discussion_r433734399) is nonsensical. `MatCheckNullSpaceMat` revealed that as well. I think I will just leave a TODO again for now.

src/mat/impls/inv/matinv.c
haplav: Again, this is caused by the current PETSc design which is obviously limiting. We would need a way to pass settings to the MUMPS factor before the actual factorization which is a legitimate request. I can think of several ways of doing this but please let's not solve it here.

src/mat/impls/inv/matinv.c
haplav: 1. I agree there is usually no point but that's out of scope of this function. This function shouldn't be responsible for deciding whether it is redundant or not. If a caller is concerned about that, it can check whether there is R already. Just like `MatGetFactor` recomputes the factor every time and it is PC responsibility to avoid redundant recomputations. 
2. Benchmarking - one can call it in the loop to get an average walltime spent in it.
3. Mat is mutable. Consider
    ```
    MatScale(K,2.0);
    MatComputeNullSpaceMat(K, NULL, MAT_ORTH_GS, MAT_ORTH_FORM_EXPLICIT, &R);
    MatSetNullSpaceMat(K,R);
    ```
    Your way R wouldn't be recomputed.

haplav:
@jkruzik The latest commit 09c7296 generalizes use of `MatCheckNullSpaceMat` by avoiding `MatGetDiagonal`. Your feedback please.

src/mat/impls/inv/matinv.c
jkruzik: 1) MatGetFactor does not actually compute the factorization. Moreover, the factor matrix is not stored in Mat. If it was, I am pretty sure it would be returned without recreation. See, e.g. [MatIsHermitian](https://www.mcs.anl.gov/petsc/petsc-current/src/mat/interface/matrix.c.html#MatIsHermitian) where the property, if known, is not recomputed.
2) If you need to run this in a loop to get a consitent timing then there is no point to benchmark this anyway as the time is minuscule.
3) That is a good point, although scaling doesn't change the null space.  In fact, changing the matrix in a certain way should invalidate the nullspace matrix. There is a provision for this (although it issues only a warning) in [MatAssemblyEnd](https://www.mcs.anl.gov/petsc/petsc-current/src/mat/interface/matrix.c.html#MatAssemblyEnd).

haplav: 1. 
    1. OK, wrong wording. MatGetFactor returns a new matrix suitable for factorization every time. So no computation but it returns a new special matrix every time. Not the best example, though.
    2. I think comparing this to MatIsHermitian is wrong because it just returns a single boolean property and has no "Compute" in its name. To mimic the same semantics, our MatGetNullSpaceMat would compute the nullspace if it is not yet computed - but that would be a bad idea e.g. because there are multiple ways and parameters for computing nullspace.
You should compare this to things like MatComputeOperator,  MatComputeBandwidth, KSPComputeEigenvalues, KSPComputeRitz, SNESComputeJacobian, ...
2. 
    1. If it is so minuscule then why you are concerned about this function doing the computation every time ;-)
    2. Everything which includes any computation is possibly subject to benchmarking. And I meant also some 3rd party benchmarking - imagine someone from outside wants to measure the time so that they know how much it takes for different matrices. They could get completely wrong times if they don't know the matrix is computed only for the first time.
    3. MUMPS is currently the only method to compute null space but in theory there can definitely be also different methods. Then it would be natural to be able to compare the times and results.
    4. If the time is rather small, the more so it should be measured in a loop to get consistent time.
    5. However, the time depends on the problem size of course. Now that we take distributed matrices into play, it is theoretically interesting to see how much time it takes to compute null space of a very big distributed matrix which for instance doesn't fit into a single node memory.
3. OK, I agree changing the matrix should invalidate the nullspace matrix. One idea: MatGetNullSpaceMat could detect whether the parent Mat has changed since the nullspace was set using PetscObjectState, and return NULL if it is the case (or throw an error)?
4. I want to keep R as output argument and then it makes absolutely no sense to _not_ recompute it. So this discussion is related to the discussion [above](#discussion_r433752090) and that one should be resolved first.

haplav: >  If it was, I am pretty sure it would be returned without recreation. See, e.g. MatIsHermitian where the property, if known, is not recomputed.

1. ii. I think comparing this to MatIsHermitian is wrong because it just returns a single boolean property and has no "Compute" in its name. To mimic the same semantics, our MatGetNullSpaceMat would compute the nullspace if it is not yet computed - but that would be a bad idea e.g. because there are multiple ways and parameters for computing nullspace.
You should compare this to things like MatComputeOperator,  MatComputeBandwidth, KSPComputeEigenvalues, KSPComputeRitz, SNESComputeJacobian, ...

jkruzik: 1) These are still bad examples. Neither function you mentioned recomputes a variable that is stored in the input object.
2) i. and iv. I clearly stated, that there is no point in benchmarking something that does take so little time that to properly benchmark it, you need to run it in a loop. Micro optimizing such things, in our case, is a waste of time. For normal size problems, the nullspace computation will take a reasonable chunk of time and that is why I do not want to recompute R if it is valid.
iii. Agreed. MatComputeNullSpaceMat should certainly be logged.
3) The problem is that not every change invalidates the null space (e.g., MatScale).
4) What is wrong with
```
MatGetNullspaceMat(mat,&Rnew);
if (Rnew) {
	*R = Rnew;
	return(0);
}
```

My point is that I see no good reason, apart from 3, why not to reuse information that we already have.

haplav: 1. It recomputes a variable stored inside only if we accept your solution where the result is directly stored into the parent Mat, which I still think is a [bad idea](#discussion_r433752090). If the user has to explicitly set it to the parent Mat like I suggest, they Compute is exactly like any other Compute function in PETSc (see examples I mentioned), the Mat is purely input argument, and the task of storing the result into the parent Mat is done explicitly and separately by the users and they know what they are doing without any unseen side effects.
2. Your arguments are based on some a posteriori knowledge of the timing. Moreover, even if the computation takes fairly long (say seconds or tens of seconds), one might want to run the loop and make average. And again, being able to do so is even more important if they want to compare different methods.
3. Actually, I wouldn't bother except of a note in docs. One should be aware that if the matrix has been changed, they might need to recompute the nullspace, but it's not the case of e.g. MatScale. OK, but that's +1 for the MatComputeNullSpaceMat recomputing always. I don't want to take the responsibility to distinguish when it should recompute and when not.

> My point is that I see no good reason, apart from 3, why not to reuse information that we already have.

But what's your legit situation where it should be reused?

If a user calls MatComputeNullSpaceMat _accidentally_ twice, it's their fault. It's not fault of this function and it shouldn't care. But at least they would see in the -log_view output it was called twice.

If the user calls MatComputeNullSpaceMat twice  _deliberately_ from _any_ reason, we shouldn't block them from doing so.

So I don't see a good reason _to reuse_.

haplav: What if I just renamed it to MatCreateNullSpaceMat which stresses a new Mat is created?

Not completely comparable but there is already [MatNullSpaceCreateRigidBody](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Mat/MatNullSpaceCreateRigidBody.html)

jkruzik: 1. You are mixing the issue with returning R with the recomputation of R. As it is now, the function recomputes R always when it is set (regardless how R was obtained) and it is useless if K did not change.
2. I just do not think that a possible use case of running a function in a loop to benchmark it, is a valid argument for recomputing a known property. On the off chance you want to do it, you can always run
```
for ...
	MatNullSpaceSetMat(mat,NULL)
	MatComputeNullSpaceMat(mat,...)
	// possibly MatDestroy(&R) if R is returned and you do not malloc Mat array for the outputs
```
3. If the common usage is
```
GetNullSpace(...,&R)
if (!R) ComputeNullSpace()
```
then it does exactly the same as if you just called `ComputeNullSpace` with what I proposed. If I change the matrix in such a way that R changes, I still need to do `SetNullSpace(...,NULL)`. (this should be mentioned in the documentation)

> But what's your legit situation where it should be reused?
>
This is a weird question. The only legit situation to recompute stuff is if either you need to because what you have is the wrong or it gains you performance (slow access to main memory, network (e.g., reduction),...). In this case, always recomputing may cover the first situation (note 3.) it will never cover 2.


I think it would be better to return R if available, but since this is more advanced/developer function, I don't, hugely, mind keeping this as is.

jkruzik: > What if I just renamed it to MatCreateNullSpaceMat which stresses a new Mat is created?
> 
Naming never was the issue. I think both names are fine. Compute may slightly better express what the function actually does.

src/mat/impls/inv/matinv.c
jkruzik: There exist a good use case. The R returned by MUMPS is dense so in a memory constrained situation I may not want to keep it in K and destroy it as soon as I do not need it.

src/mat/impls/inv/matinv.c
haplav: 1. I am mixing it because if MatComputeNullSpaceMat just computes and returns R as I propose, it's even more apparent it shouldn't just return the stashed one.
2. > If I change the matrix in such a way that R changes, I still need to do SetNullSpace(...,NULL)

Terrible. Assume I'm interested in R after changing the matrix. Let's dramatize it a bit :-) I am a "normal user" interested actually in how the R changes when the matrix changes. And write this:
```
MatCreate(comm,&R);
...
MatNullSpaceSetMat(mat,R);
//change mat values
MatComputeNullSpaceMat(mat); // does nothing
MatNullSpaceGetMat(mar,&R); // returns the same R
```
After some time, I begin to be suspicious that the R is the same. After long hours of debugging and noticing the tiny note in the docs, I "fix" it using
```
MatCreate(comm,&R);
...
MatNullSpaceSetMat(mat,R);
//change mat values
MatNullSpaceSetMat(mat,NULL); // no obvious connection to R; R is dangling
MatComputeNullSpaceMat(mat); // no obvious connection to R
MatNullSpaceGetMat(mar,&R); // only now I actually get what I'm interested in?
...
MatDestroy(&R); // destroying the new R
```
Oh, a few days later I run with valgrind and notice that the old R is actually leaked. But where should I destroy it? And then I finally fix it:
```
MatCreate(comm,&R);
...
MatNullSpaceSetMat(mat,R);
//change mat values
MatDestroy(&R);
MatNullSpaceSetMat(mat,NULL); // no obvious connection to R
MatComputeNullSpaceMat(mat); // no obvious connection to R
MatNullSpaceGetMat(mar,&R); // only now I actually get what I'm interested in?
...
MatDestroy(&R); // destroying the new R
```

For me, this is terrible API experience, and the following is much clearer and less error-prone:
```
MatCreate(comm,&R);
...
MatNullSpaceSetMat(mat,R);
//change mat values
MatDestroy(&R);
MatComputeNullSpaceMat(mat,&R); //oh, obviously new R is created so above I destroyed the old one
MatNullSpaceSetMat(mat,R);
...
MatDestroy(&R); //I know this is safe
```

> This is a weird question. The only legit situation to recompute stuff is if either you need to because what you have is the wrong or it gains you performance

I don't think it's a weird question. I want to know your use case when your stashing MatComputeNullSpaceMat saves some time.

We speak here about a function which would need to be explicitly called twice to really recompute stuff. We shouldn't protect users from such mistakes (if it was a mistake).

So I think your excuse for the unclarity demonstrated above is that you want to prevent redundant computation just in case MatComputeNullSpaceMat is called twice accidentally (which is erroneous anyway):
```
MatCreate(comm,&R);
...
MatNullSpaceSetMat(mat,R);
//change mat values
MatDestroy(&R);
MatNullSpaceSetMat(mat,NULL); // no obvious connection to R
MatComputeNullSpaceMat(mat); // no obvious connection to R
MatNullSpaceGetMat(mar,&R); // only now I actually get what I'm interested in?
...
MatComputeNullSpaceMat(mat); // forgotten by mistake, luckily no-op
...
MatDestroy(&R); // destroying the new R
```
Correct me if you have some other legit use case of calling MatComputeNullSpaceMat multiple times without any effect.

jkruzik: 1. I already [said](https://github.com/permon/permon/pull/47#discussion_r434540535), we could keep MatComputeNullSpaceMat(mat,&R) with the functionality I proposed.
2. I would suggest that user debugging something for hours that is clearly stated in the docs is a problem with the user not reading the docs in the first place.
3. `no obvious connection to R` - the function tells the object Mat to compute, if necessary, its nullspace = R
4. Valgrind argument is invalid; it has nothing to do with the proposed change. It has to do with not knowing that Seter functions typically increase ref count.
5. Agree with terrible API.
6. I do not anticipate that a user calls the function twice in the same scope. What I am thinking about is passing Mat through the stack and the function gets called somewhere on the stack by a code that the user who created Mat may not even have control off.

My point is straightforward - let's not waste time by recomputing. I find it a bit ironic that you [argued](https://github.com/permon/permon/issues/16) that the nullspace should be the property of the Mat and now you are absolutely set against taking advantage of it.

The only problem is with mat changing in a way that the nullspace changes (and you do not like the API). Using your example, I find the following more problematic
```
MatNullSpaceSetMat(mat,R);
//change mat values
....
MatDestroy(&R);
QPSetOperator(qp,mat);
QPTDualize(qp) {
    ...
    MatNullSpaceGetMat(mat,&R)
    if (!R) MatComputeNullSpaceMat(mat,&R);
    // the nullspace is not consitent!
```
the solve then fails, and it will be difficult for us to debug. This bug is much more deeply hidden. With the current state, the only solution is to ask the user to call 
```
MatNullSpaceSetMat(mat,NULL);
```
or
```
MatNullSpaceComputeMat(mat,&R);
MatNullSpaceSetMat(mat,R);
```
after any changes to mat that affect nullspace. That is error-prone.
I would propose:
```suggestion
  PetscValidPointer(R_new,5);
  TRY( MatNullSpaceGetMat(K,&R) );
  if (R) {
    TRY( MatNullSpaceChanged(K,&flg) ); // based on K object state stored by MatNullSpaceSet(), could also incorporate a MatCheckNullSpaceMat() as a secondary check to ignore changes to K that do not change the nullspace
    if (!flg) {
        *R_new = R;
        PetscFunctionReturnI(0);
    } else {
        TRY( MatDestroy(&R);
    }
  }
  TRY(PetscObjectQueryFunction((PetscObject)K,"MatComputeNullSpaceMat_C",&func) );
  if (func) {
    TRY( (*func)(K,pc,orthType,orthForm,R_new) );
  } else {
    TRY( MatComputeNullSpaceMat_Default(K,pc,orthType,orthForm,R_new) );
  }
  TRY( MatNullSpaceSetMat(K,*R_new) ); // make sure that K has the actual nullspace stored AND check that we computed a correct nullspace
  PetscFunctionReturnI(0);
}
```

Now any call to `MatComputeNullSpaceMat()` returns the correct nullspace while at the same time doing it in the cheapest way possible. I believe this fixes all the concerns raised above.

I would still prefer not returning R_new, because we are asking the Mat object to modify its own property. But it is not a big deal and typical usage is to use R after it is computed.

jkruzik: It may be better to swap `MatNullSpaceGetMat` with `MatNullSpaceChanged`.

haplav: > I would suggest that user debugging something for hours that is clearly stated in the docs is a problem with the user not reading the docs in the first place.

1. It's always better if the API is self-explanatory and tries to avoid problems by design. That definitely brings better reputation of the code.
2. I don't think the current state of docs really encourages browsing through them 😉

> Valgrind argument is invalid; it has nothing to do with the proposed change. It has to do with not knowing that Seter functions typically increase ref count.

I don't think it's invalid. My point was more that it decreases the refcount of the old object which is not directly visible and can lead to dangling pointers. This should better be avoided. I prefer that the pointer is guaranteed to be valid between Create(...,&obj) and Destroy(...,&obj) within a single scope. I hate when a function can make pointer invalid although it's not this function's argument, i.e. the internal state completely detached from the scope. This is error-prone. I think my snippets above show it clearly.

> no obvious connection to R - the function tells the object Mat to compute, if necessary, its nullspace = R

I of course meant explicit connection to the pointer R. This connection is even highlighted by the IDE, for instance. I know I'm still operating on the same object.

> I do not anticipate that a user calls the function twice in the same scope. What I am thinking about is passing Mat through the stack and the function gets called somewhere on the stack by a code that the user who created Mat may not even have control off.

1. This is so far just hypothetical.
2. If this is the case, the higher object holding the Mat could manage the lifecycle, including perhaps storing PetscObjectId and PetscObjectState of the nullspace object.
3. But keep reading, my suggestion below solves this.

> I would still prefer not returning R_new, because we are asking the Mat object to modify its own property. But it is not a big deal and typical usage is to use R after it is computed.

Yes, this would mix setting, getting and computing in the same function. This is IMHO never a good idea.

> Now any call to MatComputeNullSpaceMat() returns the correct nullspace while at the same time doing it in the cheapest way possible. I believe this fixes all the concerns raised above.

How would `MatNullSpaceChanged` look like? If it is based just on changing `PetscObjectState` of K, it would rather be `MatChanged`. But it would just compare two `PetscObjectStates` so probably not worth a new API.

But you have already argued once that e.g. `MatScale` doesn't change nullspace. How would you account for this?

Note your suggestion is buggy. On line 195, it nullifies R and destroys the linked object but the composed pointer is left dangling.

But back to my main concern, I hate that function called Compute potentially doesn't compute anything. [Should it then be called MaybeCompute 😉] And I don't like that you are trying to be more clever than the caller although you perhaps don't have enough information. [Should we employ machine learning to decide when it should be computed and when not 😉] So I still think the decision whether compute or not shouldn't be in Compute. 

But what about postponing the decision and have a similar logic in `MatNullSpaceGetMat` but it would have a specific enum modifier which would decide the way to go
1. Just return what is there. Just return `NULL` if there's nothing.
2. Definitely recompute.
3. Compute only if not yet computed or K has changed (`PetscObjectState` comparison).
4. (Optional) We could even have other options, like recompute only if the matrix has been re-assembled. I don't say we should add it now but it's nice it's potentially extensible like this.

This would kinda move the responsibility to the caller [who needs to decide the enum modifier value]. We would have a header something like
```
typedef enum {MAT_NULLSPACEMAT_RETURN,MAT_NULLSPACEMAT_COMPUTE,MAT_NULLSPACEMAT_COMPUTE_IFCHANGED} MatNullSpaceMatReuse;
PetscErrorCode MatGetNullSpaceMat(Mat mat, MatNullSpaceMatReuse flag, Mat *R);
```
This is hell of naming, we can think of better one.

So `MatComputeNullSpaceMat` would be (optionally) called from `MatGetNullSpaceMat` and could be marked as Advanced or Developer level in docs.

This would also save some lines in the caller like `QPTDualize()`. It could just always call `MatGetNullSpaceMat` with the different modifiers.

jkruzik: > > Now any call to MatComputeNullSpaceMat() returns the correct nullspace while at the same time doing it in the cheapest way possible. I believe this fixes all the concerns raised above.
> 
> How would `MatNullSpaceChanged` look like? If it is based just on changing `PetscObjectState` of K, it would rather be `MatChanged`. But it would just compare two `PetscObjectStates` so probably not worth a new API.
> 
> But you have already argued once that e.g. `MatScale` doesn't change nullspace. How would you account for this?
>
Here is a pseudocode:
```
MatNullSpaceChanged(K,*flg):
*flg = True
if K->R:
	if R->K_objStateId != K->objStateId:
		MatCheckNullSpaceMat_private(K,...,flg) //same as MatCheckNullSpaceMat but doesn't error out
```

> Note your suggestion is buggy. On line 195, it nullifies R and destroys the linked object but the composed pointer is left dangling.
> 
I do not believe so (line 204). Moreover, swapping `MatNullSpaceGetMat` and `MatNullSpaceChanged`, as [suggested](https://github.com/permon/permon/pull/47#discussion_r435120475), would get rid of 195.

> But back to my main concern, I hate that function called Compute potentially doesn't compute anything. [Should it then be called MaybeCompute wink] And I don't like that you are trying to be more clever than the caller although you perhaps don't have enough information. [Should we employ machine learning to decide when it should be computed and when not wink] So I still think the decision whether compute or not shouldn't be in Compute.
> 
How do you live with ObjSetUp functions? These functions commonly do not set up anything because the Objs are already SetUped. It is very common in programming to clearly do/return the right thing as cheaply as possible. I do have all the information, the caller wants a nullspace to K and I return a valid nullspace. How did I get it does not matter, what matters is that it is valid. If I did it cheaply and with storing a single nullspace matrix and not two is definitely a good thing.
> But what about postponing the decision and have a similar logic in `MatNullSpaceGetMat` but it would have a specific enum modifier which would decide the way to go
> 
>     1. Just return what is there. Just return `NULL` if there's nothing.
> 
>     2. Definitely recompute.
> 
>     3. Compute only if not yet computed or K has changed (`PetscObjectState` comparison).
> 
>     4. (Optional) We could even have other options, like recompute only if the matrix has been re-assembled. I don't say we should add it now but it's nice it's potentially extensible like this.
> 
Why, except for benchmarking, would you ever wanted to use 2 instead of 3? As for 4, is there any counter or something to detect re-assembly?
> 
> This would kinda move the responsibility to the caller [who needs to decide the enum modifier value]. We would have a header something like
> 
> ```
> typedef enum {MAT_NULLSPACEMAT_RETURN,MAT_NULLSPACEMAT_COMPUTE,MAT_NULLSPACEMAT_COMPUTE_IFCHANGED} MatNullSpaceMatReuse;
> PetscErrorCode MatGetNullSpaceMat(Mat mat, MatNullSpaceMatReuse flag, Mat *R);
> ```
> 
> This is hell of naming, we can think of better one.
> 
> So `MatComputeNullSpaceMat` would be (optionally) called from `MatGetNullSpaceMat` and could be marked as Advanced or Developer level in docs.
> 
> This would also save some lines in the caller like `QPTDualize()`. It could just always call `MatGetNullSpaceMat` with the different modifiers.

It does solve the problem I highlighted in the previous reply but at the cost of making the interface a bit more complicated. I'll think about this. In the meantime, please answer the above two questions.

haplav: > As for 4, is there any counter or something to detect re-assembly?

https://gitlab.com/petsc/petsc/-/blob/master/src/mat/interface/matrix.c#L5243

haplav: > It does solve the problem I highlighted in the previous reply but at the cost of making the interface a bit more complicated.

It's just one more argument and it is not more complicated for users at least, because they will not have to deal with Compute at all, and usually also not with Set.

haplav: > Why, except for benchmarking, would you ever wanted to use 2 instead of 3?

I don't say it would be _usually_ used but I have quite a few potential reasons in my mind:
1. I still think benchmarking is valid reason
    * including changing the way nullspace is computed and compare the performance _easily_
2. You may want to change the parameters of the computation from any reason
    * e.g. it could yield more accurate R after the scaling (although it doesn't change the nullspace in _exact arithmetic_)
3. Most importantly perhaps, you may want to skip the check in your `MatCheckNullSpaceMat_private` because you _know_ it will yield true.

haplav: > How do you live with ObjSetUp functions?

Well, I don't deem it comparable.
1. SetUp behaves like that consistently across PETSc and we all are used to the fact SetUp functions _guarantee_ the object is set up.
2. By contrast, all Compute functions across PETSc are stateless and compute the result anew unconditionally. Here I don't take one of your previous arguments that NullSpace is a property of the Mat and that makes difference. To me not, because it can be set as a property in a separate step (Set). If anybody is interested _only_ in the resulting nullspace (for example in `src/tests/ex3.c` here, calling Set is optional), they don't even need to know it can be stored in the Mat.
3. SetUp has only one argument which is the object being set up. But here we are speaking about a separate input matrix K and output matrix R. R is not an obligatory part of K.

jkruzik: > > As for 4, is there any counter or something to detect re-assembly?
> 
> [gitlab.com/petsc/petsc/-/blob/master/src/mat/interface/matrix.c#L5243](https://gitlab.com/petsc/petsc/-/blob/master/src/mat/interface/matrix.c#L5243)

can't detect 3rd+ reassembly

> > Why, except for benchmarking, would you ever wanted to use 2 instead of 3?
> 
> I don't say it would be _usually_ used but I have quite a few potential reasons in my mind:
> 
>     1. I still think benchmarking is valid reason
>        
>        * including changing the way nullspace is computed and compare the performance _easily_
> 
In itself, this is not worth a function argument. If you are going to write a for loop for benchmarking, you can also force `Compute` to recompute, e.g. with ObjStateIncrease.
>     2. You may want to change the parameters of the computation from any reason
>        
>        * e.g. it could yield more accurate R after the scaling (although it doesn't change the nullspace in _exact arithmetic_)
> 
>     3. Most importantly perhaps, you may want to skip the check in your `MatCheckNullSpaceMat_private` because you _know_ it will yield true.

I do not think that 2. is a good idea (as you can significantly change the solution if you push some of the eigenvalues to FP zero), but in any case, we could do a recompute based only on the ObjState.


> > How do you live with ObjSetUp functions?
> 
> Well, I don't deem it comparable.
> 
Semantically, it is the same thing -> tell an object to modify itself.
>     1. SetUp behaves like that consistently across PETSc and we all are used to the fact SetUp functions _guarantee_ the object is set up.
> 
We can guarantee, that the nullspace attached to K is correct.
>     2. By contrast, all Compute functions across PETSc are stateless and compute the result anew unconditionally. 
>
Just because PETSc doesn't do it for !Compute! functions doesn't mean that we can't do it when there is a good reason. Basically, you are saying we can't do it because it is named `MatNullSpaceComputeMat` but if the name was `MatNullSpaceSetUpMat` it would be alright...

> Here I don't take one of your previous arguments that NullSpace is a property of the Mat and that makes difference.

It is [your](https://github.com/permon/permon/issues/16) argument all along. Even this PR is named `move nullspace to mat`.
>To me not, because it can be set as a property in a separate step (Set). 

What? You just argued that is not a property of Mat.

>If anybody is interested _only_ in the resulting nullspace (for example in `src/tests/ex3.c` here, calling Set is optional), they don't even need to know it can be stored in the Mat.
> 
We are talking about recomputation and obviously, if there is no nullspace attached to Mat, then any call to `Compute()` does the computation.
>     3. SetUp has only one argument which is the object being set up. But here we are speaking about a separate input matrix K and output matrix R. R is not an obligatory part of K.
See below.

After some thought about `MatNullSpaceMatReuse` argument, I do want to have a simple `Get(mat,*mat)`. I think it is unnecessary to add the argument that potentially mixes a simple Get with Compute. Any place such an argument could go is the `Compute` (even accepting PETSC_DEFAULT) but again, if the only reason for this argument is running a benchmarking for loop, then it is not worth it.

So there remains the problem of how to solve [this example](https://github.com/permon/permon/pull/47/files#r435119281). Can you recap what is wrong with the following (I removed R as the function argument):
```suggestion
  TRY( MatNullSpaceChanged(K,&flg) ); // based on K object state stored by MatNullSpaceSet(), OPTIONALY could also incorporate a MatCheckNullSpaceMat() as a secondary check to ignore changes to K that do not change the nullspace
    if (!flg) {
        PetscFunctionReturnI(0);
  }
  TRY(PetscObjectQueryFunction((PetscObject)K,"MatComputeNullSpaceMat_C",&func) );
  if (func) {
    TRY( (*func)(K,pc,orthType,orthForm,&R) );
  } else {
    TRY( MatComputeNullSpaceMat_Default(K,pc,orthType,orthForm,&R) );
  }
  TRY( MatNullSpaceSetMat(K,R) ); // make sure that K has the actual nullspace stored AND check that we computed a correct nullspace
  TRY( MatDestroy(&R) );
  PetscFunctionReturnI(0);
}
  
```

src/mat/interface/permonmatnullspace.c
```diff
@@ -0,0 +1,69 @@
...
+  PetscFunctionReturn(0);
+}
+
+PetscErrorCode MatCheckNullSpaceMat(Mat K,Mat R,PetscReal tol)
+{
+  Vec Rx,x,y;
+  PetscReal normy;
+
+  PetscFunctionBegin;
+  if (tol == PETSC_DECIDE || tol == PETSC_DEFAULT) tol = 1e2*PETSC_SQRT_MACHINE_EPSILON;
```
jkruzik:
```suggestion
  if (tol == PETSC_DECIDE || tol == PETSC_DEFAULT) tol = 1e2*PETSC_MACHINE_EPSILON;
```

src/mat/impls/inv/matinv.c
haplav: > can't detect 3rd+ reassembly

Fair.

But note that with your overall philosophy here, `MatAssemblyBegin/End` should be no-op if the matrix hasn't changed since the last call.

> In itself, this is not worth a function argument. If you are going to write a for loop for benchmarking, you can also force Compute to recompute, e.g. with ObjStateIncrease.

We could discuss like this _indefinitely_. I say it is worth, you say it is not worth. What's the point. I am a sample user and I say for me it is worth :-D

> you can also force Compute to recompute, e.g. with ObjStateIncrease.

Awkward. Unintuitive. I complain.

> I do not think that 2. is a good idea (as you can significantly change the solution if you push some of the eigenvalues to FP zero), but in any case, we could do a recompute based only on the ObjState.

That was just an example. I don't say scaling the matrix in such a way is a good idea. _But my point you shouldn't preclude what are the user's reasons to recompute it!_ It's actually the same as `MatAssemblyBegin/End` except that R is even a separate object.

If I am a user and call Compute, I just anticipate that the computation will take place! It is _frustrating_ that I order my Mat to do something and it _doesn't do anything_. I am the boss here, I ordered compute and the stupid Mat doesn't obey :-D

The more so you cannot be 100% sure the _current_ nullspace basis has at least the same accuracy or other desired properties.

> Semantically, it is the same thing -> tell an object to modify itself.

No! It reads `MatComputeNullSpaceMat` = `Mat`, `Compute` the `NullSpace` `Mat`. I.e. order the matrix to compute another matrix which is then returned!

> We can guarantee, that the nullspace attached to K is correct.

We cannot guarantee it is correct as already discussed. In `--with-debugging=0` mode, Check is not even called automatically.

> Just because PETSc doesn't do it for !Compute! functions doesn't mean that we can't do it when there is a good reason.

I don't see your good reason, sorry.

> Basically, you are saying we can't do it because it is named MatNullSpaceComputeMat but if the name was MatNullSpaceSetUpMat it would be alright...

Yes, except we are speaking about `MatComputeNullSpaceMat` and `MatSetUpNullSpaceMat`. The latter would reflect what you propose for Compute. But it definitely shouldn't return R.

**OK, so another alternative is having both `MatComputeNullSpaceMat` and `MatSetUpNullSpaceMat`. The former always computes the NullSpaceMat, the latter guarantees it is present as property; it calls `MatComputeNullSpaceMat` if there is no NullSpaceMat yet. I think having these two simple functions is a fair compromise.**

>> Here I don't take one of your previous arguments that NullSpace is a property of the Mat and that makes difference.
>
> It is your argument all along. Even this PR is named move nullspace to mat.
>
>> To me not, because it can be set as a property in a separate step (Set).
>
> What? You just argued that is not a property of Mat.
>

This is misunderstanding. I don't question whether NullSpaceMat is a property, but that it's kind of unrelated. We should distinguish between a class property and a (runtime) object property. The aim of my Compute function is to compute Mat instance R from Mat instance A, so that R approximates the nullspace basis of A and returns it. R becomes a NullSpaceMat property of A once it is attached to it _intentionally_ using Set. These are for me _two separate steps_. 

For example, I could have A1 and A2 that are different matrices that _mathematically_ have the same nullspace (or similar up to a given tolerance). They could be copies, or A1 approximating A2 (e.g. A1 being an explicit approximation of implicit A2), or they could even have different values. Then I could use Compute to compute R from A1 but attach it to A2. So R would be a _runtime property_ of A2 although computed from A1. And then I could even immediately change A1 so that R is no longer an approximation of nullspace of A1.

> After some thought about MatNullSpaceMatReuse argument, I do want to have a simple Get(mat,*mat). I think it is unnecessary to add the argument that potentially mixes a simple Get with Compute. 

OK, I agree that Get/Set functions should arguably be kept simple and without any heavy computation.

> So there remains the problem of how to solve this example. Can you recap what is wrong with the following (I removed R as the function argument):

Nothing, except the Compute function shouldn't behave like this :-D Some automagic under the hood - as argued multiple times - no point in repeating myself.

**So what about having both Compute and SetUp as detailed above in bold font?**

jkruzik: Alright, we probably can't agree on this. In the interest of moving forward here: I am for the SetUp function or perhaps something like MatNullSpaceMatIsValid() which can be used in QPTDualize to check if the null space attached to the mat is valid or needs to be recomputed. In either case, the validity check may suffice to be on ObjState of Mat (K).

Note that, I still think there should be MatCheckNullSpaceMat in dbg mode in Compute function.

src/mat/interface/permonmatnullspace.c
haplav: This has already been discussed in https://github.com/permon/permon/commit/09c7296b7a2f8b8a2e6860182dbdb28ba8824848#r39688564

>>> Let's test the absolute criterium - it should be always correct. I would suggest the default tolerance to be 100*eps.
>>
>> Do you mean eps = PETSC_SQRT_MACHINE_EPSILON? That's what MatNullSpaceTest uses.
>
>yes

See also https://gitlab.com/petsc/petsc/-/blob/master/src/mat/interface/matnull.c#L409

jkruzik: OK, my mistake in the previous discussion. I meant eps = PETSC_MACHINE_EPSILON. Using 1.49e-5 in double is extremely weak.

haplav: Where did you dig this number?

https://gitlab.com/petsc/petsc/-/blob/master/include/petscmath.h#L777

jkruzik: *e-6, sorry

haplav: Got you, you already multiplied by the given factor but still it is 1.49e-6, right?

haplav: I'm reluctant to use stronger criterion than `MatNullSpaceTest` since we even throw an error.

src/mat/impls/inv/matinv.c
haplav: > I am for the SetUp function or perhaps something like MatNullSpaceMatIsValid() which can be used in QPTDualize to check if the null space attached to the mat is valid or needs to be recomputed. In either case, the validity check may suffice to be on ObjState of Mat (K).

OK, let me put it off to #20 where we should decide finally about interplay of QPTDualize and NullSpaceMat.

haplav:
https://github.com/permon/permon/pull/47#discussion_r441477104
> Note that, I still think there should be MatCheckNullSpaceMat in dbg mode in Compute function.

OK, done now in fb2f271.

Note also `MatCheckNullSpaceMat` now returns flag instead of failing. That gives more flexibility how it can be used.

haplav:
(partial squash)

src/mat/impls/inv/matinv.c
jkruzik: If we are going to track K obj state for checking R, it should be done here. If it is not done, there is a bug as shown [here](https://github.com/permon/permon/pull/47#discussion_r435119281).

src/mat/interface/permonmatnullspace.c
jkruzik: Do not be so fixated on what is done in PETSc. The important thing is why should we use such a weak tolerance. Using 1.49e-6 allows for perturbations in the nullspace of order ~1e-6 (i.e. 3x3 Laplacian with nullspace [1+1e-6;1;1] satisfies the tolerance). I think that 1e-8 (sqrt_eps) should be an absolute minimum for the tolerance (which coincidently, I believe, is the threshold for pivots when computing the nullspace).

haplav: > Do not be so fixated on what is done in PETSc. 

I am, because my overall plan this functionality should be in PETSc anyway. If I retrospect how much time I spent with this PERMON PR, I think I should have done that right away.

> Using 1.49e-6 allows for perturbations in the nullspace of order ~1e-6 (i.e. 3x3 Laplacian with nullspace [1+1e-6;1;1] satisfies the tolerance). I think that 1e-8 (sqrt_eps) should be an absolute minimum for the tolerance (which coincidently, I believe, is the threshold for pivots when computing the nullspace).

PETSc has `10*sqrt_eps` ~ 1e-7, not `100*sqrt_eps` [which I originally thought you were proposing]. I'm OK with sqrt_eps.

src/mat/impls/inv/matinv.c
haplav: Can you be more specific about the bug? Can you please comment directly in the related code?

src/mat/interface/permonmatnullspace.c
haplav: Changed to sqrt_eps in 705fc45. Closing.

haplav:
TODO
* [ ] Move nullspace computation from [`matinv.c`](https://github.com/permon/permon/pull/47/files#diff-99556492510148a43536b8fdd4e88e24R53) to `permonmatnullspace.c`
* [ ] Add manpages to functions in `permonmatnullspace.c`
* [ ] Do some more squashing.

src/qp/interface/qptransform.c
```diff
@@ -981,19 +983,20 @@ PetscErrorCode QPTDualize(QP qp,MatInvType invType,MatRegularizationType regType
 
   /* get or compute stiffness matrix kernel (R) */
   R = NULL;
-  TRY( QPGetOperatorNullSpace(qp,&R) );
-  if (R) {
-    TRY( MatInvSetNullSpace(Kplus,R) );
-  } else {
+  TRY( MatGetNullSpaceMat(K,&R) );
+  if (!R) {
```
jkruzik, line 987: Can lead to a wrong result, if K changes after a call to MatSetNullSpaceMat(). We need to ensure that R is a valid nullspace as I was saying throughout https://github.com/permon/permon/pull/47#discussion_r442102721. My suggestion is to store K ObjState on Set().

haplav: I agree this is an issue (quite hypothetical at the moment, though).

However, as you can see in diff, view, this issue is not introduced by this PR. It was there before, and it's not a goal of this PR to solve it. It's more convenient for me to fix everything about nullspace use in QPTDualize in #20 and fits its scope better. And there will probably be another discussion how exactly all this will be implemented.

haplav: mentioned now in TODOs https://github.com/permon/permon/pull/20#issuecomment-635186807

jkruzik: Though, the bug is introduced by this PR. We had QPSetNullSpaceOperator() and therefore users knew that QP object uses R - changing K after setting R to QP is a bug in user code. Now, R is hidden in K and users have no idea that QP uses R.

Almost all of the comments from my point of view in https://github.com/permon/permon/pull/47#discussion_r442102721 were about solving this bug regardless of superficial naming and benchmarking stuff. It should be solved here as it is not related to the singular Hessian in QPTDualize.

As was pointed out many times, there is an easy solution with ObjState tracking.

haplav: > Though, the bug is introduced by this PR.

This is funny. First of all, the "bug" is not even covered by tests. So it's rather hypothetical. Feel free to add a few tests in #20 which will make this bug reproducible and I will have to fix it in order to make tests passing.

> We had QPSetNullSpaceOperator() and therefore users knew that QP object uses R - changing K after setting R to QP is a bug in user code. Now, R is hidden in K and users have no idea that QP uses R.

If the user has no idea, then she/he probably doesn't set the nullspace at all and then it gets computed automatically in QPTDualize with correct results. _No change here._

If additionally they change K after QPTDualize and then call QPTDualize again, they get wrong results. _No change here._ [Quite unlikely use case as of now, though.]

It the user explicitly sets/computes nullspace, it is intentional act on their side and they take the responsibility. _No change here._

And if they change the operator afterwards, then they get wrong results. _No change here._

There was never before a mechanism to really protect from these issues. Stating that it had been a user code bug and now it is our bug, just because R is now directly a K property, is ridiculous. And it's the more so fishy argumentation that it has never been properly explained in docs so it's always our fault actually.

However, there could have been problems with syncing R in MATINV and R in QP before, which are now gone at least.

> As was pointed out many times, there is an easy solution with ObjState tracking.

I don't question it. But it's not the only solution. I asked you to postpone the discussion to #20 and you once again make me to add some reasons for it, which is quite frustrating, but OK.

We could as well make really sure the provided nullspace is correct by calling CheckNullSpaceMat in QPTDualize - its cost is a few purely local matrix-vector products and maybe it would pay off to do that rather than rely on ObjState or so. I don't advocate that right now and I don't wanna discuss it here. I need to think about that. And it's a matter of #20 which will follow.

Could you please let me do things in my preferred order since I'm the one who will implement it eventually? I promised I would solve it in #20. I thought that should normally suffice. So if you let me do my job, I will finish the [TODOs](https://github.com/permon/permon/pull/47#issuecomment-645911900) and it will be ready for merge.

jkruzik: I probably did not make myself clear, **what I want to know is why do you believe it is better to fix it in #20 and not here**. My reasoning is that it clearly falls into the scope of this PR while it has nearly nothing to do with non-singular Hessian in QPTDualize. The other thing is that merging something with a known fault with a view of hopefully fixing it down the line is a bad idea. To reiterate, it is possible to solve it in #20 if there is a good reason to do it there and it falls into its scope (I really struggle to see the latter point).

Another solution, which would suffice and is the simplest, is to leave this as is, but then it should be mentioned in the doc to QP object and QPTDualize that it uses R attached using MatSetNullSpaceMat() to the operator set by QPSetOperator().

At first, I wanted to end my reply here and skip replying to most of the things you raised because it is actually not related to what I need to know from you. Clearly, whether this problem stems from the changes in this PR (which has nothing to do with a simple diff but rather the composition of the objects) can be contested, but ultimately the problem is here, it was found due to this PR and is clearly related to it. However, I do feel obligated to clear some of the points and comment on some of your reasoning that I considered flawed.

> This is funny. First of all, the "bug" is not even covered by tests. So it's rather hypothetical.
>
Saying that a problem is not covered by a test is irrelevant - it is still a problem
> Feel free to add a few tests in #20 which will make this bug reproducible and I will have to fix it in order to make tests passing.
> 
I am not going to put code directly into your PR unless you specifically ask me to do so. However, if I were to do that I could choose to add the aforementioned test into this PR.
> > We had QPSetNullSpaceOperator() and therefore users knew that QP object uses R - changing K after setting R to QP is a bug in user code. Now, R is hidden in K and users have no idea that QP uses R.
> 
> If the user has no idea, then she/he probably doesn't set the nullspace at all and then it gets computed automatically in QPTDualize with correct results. _No change here._
> 
I said that, with these changes, the user does not know that setting null space into K affects operations on QP object. There is a fundamental difference between
```QPSetOperatorNullSpace(QP,Mat)```
which operates on QP objects and
```MatSetNullSpaceMat(Mat,Mat)```,
because in the latter there is no obvious connection to QP. That should certainly be mentioned in the doc.
> If additionally they change K after QPTDualize and then call QPTDualize again, they get wrong results. _No change here._ [Quite unlikely use case as of now, though.]
> 
I never mentioned this anywhere.
> It the user explicitly sets/computes nullspace, it is intentional act on their side and they take the responsibility. _No change here._
> 
> There was never before a mechanism to really protect from these issues. Stating that it had been a user code bug and now it is our bug, just because R is now directly a K property, is ridiculous. And it's the more so fishy argumentation that it has never been properly explained in docs so it's always our fault actually.
> 
Again the difference is that with setting the nullspace to Mat there is no obvious connection to QP.
> However, there could have been problems with syncing R in MATINV and R in QP before, which are now gone at least.
> 
Yes, that is good although besides the point.
> > As was pointed out many times, there is an easy solution with ObjState tracking.
> 
> I don't question it. But it's not the only solution. I asked you to postpone the discussion to #20 and you once again make me to add some reasons for it, which is quite frustrating, but OK.
> 
Yes, I did want some good reasons to postpone this to #20, and as a reply, I did get a very long message that does not give any reason to postpone it. You just argued why this problem is _similar_ to the state before the change.
>  I promised I would solve it in #20. I thought that should normally suffice. 
>
No. What suffices is a clear reason why it should be done in #20.

PR #49, opened 2021-04-06
fix compilation with C++ compiler

jkruzik:
PETSc configured with --with-clanguage=c++

Reported-by: Zach Langford zlangfor@vols.utk.edu

### DIFF
--- a/include/permonmat.h
+++ b/include/permonmat.h
@@ -151,6 +151,7 @@ FLLOP_EXTERN PetscErrorCode MatRedistributeRows(Mat mat_from,IS rowperm,PetscInt
 
 /* FETI UTILITIES */
 FLLOP_EXTERN PetscErrorCode MatRemoveGluingOfDirichletDofs(Mat Bgt, Vec cg, Mat Bdt, Mat *Bgt_new, Vec *cg_new, IS *is_new);
+FLLOP_EXTERN PetscErrorCode MatRemoveGluingOfDirichletDofs_old(Mat Bgt, Vec cg, Mat Bdt, Mat *Bgt_new, Vec *cg_new, IS *is_new);
 
 /*   ORTHONORMALIZATION   */
 typedef enum {MAT_ORTH_NONE=0, MAT_ORTH_GS, MAT_ORTH_GS_LINGEN, MAT_ORTH_CHOLESKY, MAT_ORTH_IMPLICIT, MAT_ORTH_INEXACT} MatOrthType;
--- a/src/mat/interface/permonmatregularize.c
+++ b/src/mat/interface/permonmatregularize.c
@@ -201,14 +201,15 @@ static PetscErrorCode MatRegularize_GetRegularization_Private(Mat K_loc, Mat R_l
 #undef __FUNCT__  
 #define __FUNCT__ "MatRegularize"
 PetscErrorCode MatRegularize(Mat K, Mat R, MatRegularizationType type, Mat *newKreg) {
-  static PetscBool  registered = PETSC_FALSE;
-  static PetscInt   regularized_id;
-  MPI_Comm          comm;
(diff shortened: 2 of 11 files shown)


src/qp/interface/qpchain.c
```diff
@@ -239,7 +239,7 @@ PetscErrorCode QPChainPostSolve(QP qp)
   }
 
   TRY( PetscOptionsGetViewer(comm,NULL,prefix,"-qp_chain_view_kkt",&v,&format,&view) );
-  view &= !PetscPreLoadingOn;
+  view = (PetscBool)(view & !PetscPreLoadingOn);
```
haplav:
```suggestion
  view = (PetscBool)(view && !PetscPreLoadingOn);
```

src/mat/interface/permonmatregularize.c
```diff
@@ -208,6 +208,7 @@ PetscErrorCode MatRegularize(Mat K, Mat R, MatRegularizationType type, Mat *newK
   Mat               Q_loc, Kreg;
   PetscScalar       rho;
   Mat               K_loc, R_loc;
+  PetscInt          regularized_int;
```
haplav:
```suggestion
  int               regularized_int;
```
as PetscObjectComposedDataGetInt() works with int

haplav: Also please fix `regularized_id` above. My fault.

src/qp/interface/qpchain.c
haplav: (my old fault, sorry)

include/permonmat.h
```diff
@@ -151,6 +151,7 @@ FLLOP_EXTERN PetscErrorCode MatRedistributeRows(Mat mat_from,IS rowperm,PetscInt
 
 /* FETI UTILITIES */
 FLLOP_EXTERN PetscErrorCode MatRemoveGluingOfDirichletDofs(Mat Bgt, Vec cg, Mat Bdt, Mat *Bgt_new, Vec *cg_new, IS *is_new);
+FLLOP_EXTERN PetscErrorCode MatRemoveGluingOfDirichletDofs_old(Mat Bgt, Vec cg, Mat Bdt, Mat *Bgt_new, Vec *cg_new, IS *is_new);
```
haplav, line 154: Could you move it to `include/permon/private/permonmatimpl.h` not to pollute public headers?

haplav (approved):

haplav:
@jkruzik Thanks for taking care of this!

mpecha (approved):

include/permonmat.h
jkruzik: I think we could merge them or remove the old one entirely. I will put up an issue.

src/mat/interface/permonmatregularize.c
jkruzik: I just checked, the `PetscInt regularized_int` is correct (documentation is wrong). The `regularized_id` is of type `PetscObjectState` which is `PetscInt64`. I'm going to change the id to the correct type and declare the `regularized_int` as `PETSC_UNUSED` to get rid of the `(void)` cast below.

jkruzik: `regularized_id` will need to be `PetscInt` due to `PetscObjectComposedDataRegister()`

include/permonmat.h
haplav: I don't remember the history here but I would just make some simple performance comparison to make sure the newer version is at least at the same speed, and then it could be deleted completely.

src/mat/interface/permonmatregularize.c
haplav: Yes, you're right, I was checking just the manpage - should be fixed in PETSc, I will do it.

PR #51, opened 2021-04-13
Support PETSc 3.15

jkruzik:


src/mat/impls/inv/matinv.c
```diff
@@ -546,33 +551,33 @@ static PetscErrorCode MatInvCreateInnerObjects_Inv(Mat imat)
...
+      inv->innerksp = *kspp;
+    } else {
+      char stri[1024];
+      TRY( PetscSNPrintf(stri, sizeof(stri), "-%smat_inv_redundant_pc_type none",prefix) );
+      TRY( PetscOptionsInsertString(NULL,stri) );
+      //petsc bug start https://bitbucket.org/petsc/petsc/branch/hzhang/fix-pcredundant#diff
+      TRY( PetscSNPrintf(stri, sizeof(stri), "-psubcomm_type %s", PetscSubcommTypes[inv->psubcommType]) );
+      TRY( PetscOptionsInsertString(NULL,stri) );
+      TRY( PCSetFromOptions(pc) );
+      //bug end
```
haplav: Could this be removed now?

src/mat/impls/inv/matinv.c
```diff
@@ -546,33 +551,33 @@ static PetscErrorCode MatInvCreateInnerObjects_Inv(Mat imat)
...
+      TRY( PCSetType(pc, PCREDUNDANT) );
+      TRY( PCRedundantSetNumber(pc, inv->redundancy) );
+      TRY( PCSetUp(pc) );//not necessary after fix, see above
```
haplav: ... and this?

jkruzik: that was already fixed in 07d0622

haplav: aha, ok, sorry - I was looking at the old commit

haplav: Wait - you just removed the comments, right? I was thinking that the comments say the ugly code is a workaround that could be revisited once that branch gets merged. It was this https://gitlab.com/petsc/petsc/-/commit/f07d178c4121a870076ccb0b7ed834dcdbc01e49

No need to deal with that right away but perhaps we could keep the comments and create an issue. I or you should take a look at it at some point.

haplav: OK, issue #53 created. No need to revert the comments since this discussion and the issue does the job. Resolving.

haplav:
OK, @jkruzik, I think this is ready for merge.

src/mat/impls/inv/matinv.c
jkruzik: I actually checked before removing the comments ;) And simply removing `PCSetUp` makes problems elsewhere. However, looking at it again I can save a few lines of code (saving a NOP `PcSetUp`). See the amended last commit.

mpecha (approved):

PR #52, opened 2021-04-13
permon_rules: update crash email address

haplav:
@jkruzik I think this is the reason Zach contacted only me in the first place 🙂


jkruzik (approved):
Thanks! I wanted to fix this but forgot.

PR #74, opened 2022-11-23
Fixes #17; Removed obsolete undef

### LINKED ISSUE #17: Get rid of #undef __FUNCT__ / #define __FUNCT__
All these
```
#undef __FUNCT__
#define __FUNCT__ "main"
```
are legacy of > 2 years old PETSc and should be removed.

Following up https://github.com/permon/permon/pull/12#pullrequestreview-401664011

davidhrbac:


src/mat/impls/petscfix/sbaij.c
```diff
@@ -2,11 +2,7 @@
 #include <permon/private/petscimpl.h>
 
 #if defined(MatCreateSeqSBAIJWithArrays)
-#undef MatCreateSeqSBAIJWithArrays
-#endif
```
jkruzik, line 6: this can't be deleted

I'll open an issue to review these fixes for PETSc functions.

jkruzik (changes requested):
`__FUNCT__` is used in some parts of the code and needs to be replaced by `PETSC_FUNCTION_NAME` which is defined as `__func__`. Therefore, printf using it like
https://github.com/permon/permon/blob/58b0ebb1d39dae515962c1234748de18def15003/src/qp/interface/qpchain.c#L396
need to replace it as
```PetscCall(PetscViewerASCIIPrintf(v,"%s output follows\n",PETSC_FUNCTION_NAME));```

For a reference, this PR removed `__FUNCT__` from PETSc https://gitlab.com/petsc/petsc/-/merge_requests/609

